PR #1, opened 2016-11-05
Initial commit

ethitter:
Moving over from https://github.com/Automattic/vip-go-mu-plugins/pull/329.

Rather than executing scheduled cron tasks in series, multiple tasks can run simultaneously, preventing long-running tasks from blocking other events. 

**How**
1. An endpoint provides a list of tasks to run in the next 60 seconds. Enough information about each is provided to identify it, but its specific arguments are not exposed.
2. A second endpoint accepts the JSON object for an individual task. After confirming that it isn't scheduled more than 15 seconds ahead of the current time, the endpoint runs that job, mimicking the environment Core provides.
3. The existing runners on WordPress.com are updated to use these two endpoints. For each cron task to run, an individual job will be queued on WordPress.com's jobs system.

_All_ attempts to execute Core's native cron methods (`wp-cron.php`, `wp_cron()`, `spawn_cron()`) are blocked to prevent rogue execution.


ethitter:
There's one major issue right now: the CPT-based storage doesn't account for unscheduling an event without running it. I'm working out how to detect that something's been removed for this reason.

ethitter:
`wp_unschedule_event()` is handled now thanks to 9f26e26, so I'm performing some final tests before requesting another review.

mjangda:
I think we should use either `Automattic\CronControl` or `Automattic\WP\CronControl` as the namespace.

ethitter:
Good call; I'm terrible at naming things. I'll update to use the latter format for clarity, and because we don't just release WP-related things.

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+	 * Create a job post, respecting whether or not Core is ready for CPTs
+	 */
+	private function create_job( $job_post ) {
+		// Limit how many events to insert at once
+		if ( ! Lock::check_lock( self::LOCK, 5 ) ) {
+			return false;
+		}
+
+		// If called before `init`, we need to insert directly because post types aren't registered earlier
+		if ( did_action( 'init' ) ) {
```
mjangda: `wp_insert_post` doesn't care if the post_type exists so we should be able to use it directly (and scrap the entire else clause there).

```
wp> wp_insert_post( [ 'post_type' => 'not-real', 'post_title' => 'title' ] )
int(1852)
wp> get_post( 1852 )->post_title
string(5) "title"
wp> get_post( 1852 )->post_type
string(8) "not-real"
```

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+	/**
+	 * Remove an event's CPT entry
+	 *
+	 * @param $timestamp  int     Unix timestamp
+	 * @param $action     string  name of action used when the event is registered (unhashed)
+	 * @param $instance   string  md5 hash of the event's arguments array, which Core uses to index the `cron` option
+	 *
+	 * @return bool
+	 */
+	private function delete_job( $timestamp, $action, $instance ) {
```
mjangda: Should the delete be checking for a lock too?

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+			'posts_per_page'   => 1,
+		) );
+
+		if ( ! $job ) {
+			return false;
+		}
+
+		// If called before `init`, we need to delete directly because post types aren't registered earlier
+		if ( did_action( 'init' ) ) {
+			wp_delete_post( $job->ID, true );
```
mjangda: `wp_delete_post` doesn't care about post_type either.

includes/class-events.php
```diff
@@ -0,0 +1,172 @@
...
+		$events = get_option( 'cron' );
+
+		// That was easy
+		if ( ! is_array( $events ) || empty( $events ) ) {
+			return array( 'events' => null, );
+		}
+
+		// To be safe, re-sort the array just as Core does when events are scheduled
+		// Ensures events are sorted chronologically
+		uksort( $events, 'strnatcasecmp' );
```
mjangda: Do we really need to re-sort since `get_option` already does it?

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+		} else {
+			global $wpdb;
+
+			$wpdb->delete( $wpdb->postmeta, array( 'post_id' => $job->ID, ) );
+			$wpdb->delete( $wpdb->posts, array( 'ID' => $job->ID, ) );
```
mjangda: What if instead of deleting during the process (which may be slow), we flag events as completed (maybe post_stutus => private) and filter them them when getting the posts? And then a separate job goes through and deletes them.

includes/constants.php
```diff
@@ -0,0 +1,19 @@
...
+const JOB_EXECUTION_BUFFER_IN_SECONDS = 15;
+const JOB_TIMEOUT_IN_MINUTES          = 10;
+const JOB_CONCURRENCY_LIMIT           = 10;
+
+/**
+ * REST API
+ */
+const REST_API_NAMESPACE     = 'cron-control/v1';
+const REST_API_ENDPOINT_LIST = 'events';
+const REST_API_ENDPOINT_RUN  = 'event';
```
mjangda: We can probably move this to REST_API class.

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+					'suppress_filters' => false,
+					'posts_per_page'   => 1,
+				) );
+
+				if ( ! $job_exists ) {
+					// Build minimum information needed to create a post
+					$job_post = array(
+						'post_title'            => $this->event_title( $event['timestamp'], $event['action'], $event['instance'] ),
+						'post_name'             => $this->event_name( $event['timestamp'], $event['action'], $event['instance'] ),
+						'post_content_filtered' => maybe_serialize( array(
```
ethitter: Since we're only using `post_content_filtered`, there's no risk. Neither `wp_insert_post()` nor `sanitize_post_field()` modify the value from that column.

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+		}
+
+		return true;
+	}
+
+	/**
+	 * Determine if current request is a call to `wp_unschedule_event()`
+	 */
+	private function is_unscheduling() {
+		return false !== array_search( 'wp_unschedule_event', wp_debug_backtrace_summary( __CLASS__, null, false ) );
```
ethitter: > Any concerns that this might be slow?

Nope, PHP's `debug_backtrace()` (the underlying call) doesn't slow things down significantly. Using `timer_start()` and `timer_stop()` to profile the call and `array_search` revealed an average execution time of 0.0000511 seconds.

includes/class-cron-options-cpt.php
ethitter: I like it. I'll change the post status to `trash` and a job can trigger the cleanup process Core uses when trashed-posts' grace period expire.

includes/class-events.php
ethitter: Core only re-orders the array when `wp_schedule_event()` is called, which probably happens frequently enough (`wp_reschedule_event()` calls it too), but I included this here just in case something directly manipulated the option. Since Core doesn't reorder when calling `_get_cron_array()`, it's probably safe for me to remove this. :)

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+		}
+	}
+
+	/**
+	 * Check if a job post exists, respecting Core's loading order
+	 */
+	private function job_exists( $job_post ) {
+		// If called before `init`, we need to insert directly because post types aren't registered earlier
+		if ( did_action( 'init' ) ) {
+			$exists = get_posts( $job_post );
```
ethitter: Good call, `get_page_by_path()` is safe to use before init, and there's some added caching too.

includes/class-cron-options-cpt.php
ethitter: Since it calls `post_type_supports()`, it should wait until after `init`: https://core.trac.wordpress.org/browser/tags/4.6.1/src/wp-includes/post.php#L2939. That said, `post_type_supports()` doesn't error when called early, and the places it's used are ones we either don't care about (thumbnails) or ones we can overcome with filters (`wp_insert_post_empty_content`). I'll just call it directly.

includes/class-cron-options-cpt.php
ethitter: A number of the functions it calls do, though looking them over, we'd be safe; we aren't using revisions, nor is the post type hierarchical. Given that we're switching to trashing the post and cleaning it up later, it's moot anyway: `wp_trash_post()` is safe to call before `init`. 🎉

includes/class-cron-options-cpt.php
ethitter: Never mind, this needs to stay:

```
[18-Nov-2016 19:16:38 UTC] PHP Notice:  Trying to get property of non-object in /srv/www/wordpress-default/wp-includes/post.php on line 3579
[18-Nov-2016 19:16:38 UTC] PHP Notice:  Trying to get property of non-object in /srv/www/wordpress-default/wp-includes/link-template.php on line 49
[18-Nov-2016 19:16:38 UTC] PHP Fatal error:  Call to undefined function is_user_logged_in() in /srv/www/wordpress-default/wp-includes/post.php on line 2066
```

includes/class-cron-options-cpt.php
```diff
@@ -0,0 +1,375 @@
...
+						'args'     => $args['args'],
+					);
+
+					if ( isset( $args['interval'] ) ) {
+						$cron_array[ $timestamp ][ $action ][ $instance ]['interval'] = $args['interval'];
+					}
+
+				}
+			}
+		} while( true );
```
ethitter: I'm undecided. The risk with not returning all of the entries is that the code that created the event will keep trying to schedule it. There's rate-limiting elsewhere, which mitigates some of this risk, but the existence check may not catch the event if the code sets a new timestamp with each rescheduling.

ethitter:
@mjangda Let me know your thoughts on the `do` loop. I've addressed the remaining items either in code or with a comment above.

includes/class-cron-options-cpt.php
mjangda: We're probably safe to break after `$page > 10`? A site with 1000+ cron entries probably has bigger problems? My reasoning is mostly to protect ourselves if a logic error in the code fails to break out causing the loop to run forever and 💥 .

includes/class-cron-options-cpt.php
```diff
@@ -308,7 +308,25 @@ private function mark_job_completed( $timestamp, $action, $instance ) {
...
-		wp_trash_post( $job->ID );
+		$this->mark_job_post_completed( $job->ID );
+
+		return true;
+	}
+
+	/**
+	 * Set a job post to the "completed" status
+	 *
+	 * `wp_trash_post()` calls `wp_insert_post()`, which can't be used before `init` due to capabilities checks
```
mjangda: Oh, WordPress... Sigh...

ethitter:
Since switching to `get_page_by_path()`, I've started seeing duplicate events on first population. I suspect that in `create_job()`, I need to bust the `last_modified` cache that `get_page_by_path()` uses as part of its key. I may also add some handling to the options CPT to detect and purge duplicates.

mjangda:
> Since switching to get_page_by_path(), I've started seeing duplicate events on first population.

Maybe we should just do our own direct query?

ethitter:
> Maybe we should just do our own direct query?

Indeed, I'm thinking I'll restore the direct query that I removed in https://github.com/Automattic/Cron-Control/pull/1/commits/2635d7aee68eddd84b1e3d42f6fc561476eaad1f#diff-b70d366c95711166090fd520fd949aa5L225, but without the switch around the `init` action. Avoiding duplicate events is worth the direct query, and it's using indexes.

includes/class-cron-options-cpt.php
ethitter: Fair enough. Added in 9c3a092.

ethitter:
Okay, ready for another look @mjangda.

mjangda:
Looking good! Only small thing: let's make the post_type slug singular to follow convention (`a8c_cron_ctrl_events` => `a8c_cron_ctrl_event`)

mjangda:
🚢  🚢  🚢

PR #32, opened 2016-12-06
Introduce CLI command to purge events

### LINKED ISSUE #27: Should plugin have CLI command to purge its CPT store
If data is corrupted, or the plugin is retired from a site, it would be convenient to have a command to remove its data.

ethitter:
From time to time, perhaps when retiring this plugin, it may be necessary to remove all of its data.

Fixes #27

### DIFF
--- a/includes/wp-cli/class-one-time-fixers.php
+++ b/includes/wp-cli/class-one-time-fixers.php
@@ -7,7 +7,7 @@ class One_Time_Fixers extends \WPCOM_VIP_CLI_Command {
 	/**
 	 * Remove corrupt Cron Control data resulting from initial plugin deployment
 	 *
-	 * eg.: `wp --allow-root cron-control remove-all-plugin-data`
+	 * eg.: `wp --allow-root cron-control remove-all-plugin-data --batch-size=15 --dry-run=true`
 	 *
 	 * @subcommand remove-all-plugin-data
 	 */
@@ -28,7 +28,7 @@ class One_Time_Fixers extends \WPCOM_VIP_CLI_Command {
 
 		if ( is_numeric( $count ) ) {
 			$count = (int) $count;
-			\WP_CLI::line( sprintf( __( 'Found %s entries', 'automattic-cron-control' ), number_format_i18n( $count ) ) . "\n\n" );
+			\WP_CLI::line( sprintf( __( 'Found %s total items', 'automattic-cron-control' ), number_format_i18n( $count ) ) . "\n\n" );
 		} else {
 			\WP_CLI::line( __( 'Something went wrong...aborting!', 'automattic-cron-control' ) );
 			return;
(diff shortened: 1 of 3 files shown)


includes/wp-cli/class-one-time-fixers.php
```diff
@@ -0,0 +1,106 @@
...
+		// Provide some idea of what's going on
+		\WP_CLI::line( __( 'CRON CONTROL', 'automattic-cron-control' ) . "\n" );
+
+		$count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(ID) FROM {$wpdb->posts} WHERE post_type = %s;", 'a8c_cron_ctrl_event' ) );
+
+		if ( is_numeric( $count ) ) {
+			$count = (int) $count;
+			\WP_CLI::line( sprintf( __( 'Found %s total items', 'automattic-cron-control' ), number_format_i18n( $count ) ) . "\n\n" );
+		} else {
+			\WP_CLI::line( __( 'Something went wrong...aborting!', 'automattic-cron-control' ) );
```
nickdaugherty: This should probably be a `\WP_CLI::error()`

includes/wp-cli/class-one-time-fixers.php
```diff
@@ -0,0 +1,106 @@
...
+			if ( ! is_array( $items ) || empty( $items ) ) {
+				\WP_CLI::line( __( 'No more items found!', 'automattic-cron-control' ) );
+				break;
+			}
+
+			\WP_CLI::line( sprintf( __( 'Found %s items in this batch' ), number_format_i18n( count( $items ) ) ) );
+
+			if ( ! $dry_run ) {
+				foreach ( $items as $item ) {
+					\WP_CLI::line( "{$item->ID}, `{$item->post_title}`" );
```
nickdaugherty: For the dry run, we should also print out each item id and title, so the runner can spot-check it for accuracy before committing to a real run.

includes/wp-cli/class-one-time-fixers.php
```diff
@@ -0,0 +1,106 @@
...
+		} while( true );
+
+		// Remove the now-stale cache when actively run
+		if ( ! $dry_run ) {
+			wp_cache_delete( 'a8c_cron_ctrl_option' );
+			\WP_CLI::line( "\n" . sprintf( __( 'Cleared the %s cache', 'automattic-cron-control' ), 'Cron Control' ) );
+		}
+
+		// Fin
+		\WP_CLI::line( "\n" . __( 'All done.', 'automattic-cron-control' ) );
```
nickdaugherty: We can also use `\WP_CLI::success()` here.

includes/wp-cli/class-one-time-fixers.php
ethitter: Heh, yes, but I was being lazy and hoped you wouldn't notice. 😉

Since the query doesn't deal with pagination, the output during a dry-run doesn't change, but I'll move this back up anyway. Some output is better than nothing.

Issue #28, opened 2016-12-05
Introduce CLI command to deal with caches, CPT, etc


ethitter:
It would be helpful if the plugin registered some common commands for dealing with its caches and data store, removing the need for developers to remember how the plugin is architected simply to support its basic operations.

ethitter:
In addition to the cache-clearing function from 977f561, the plugin should provide:

- list pending items
- list completed items
- delete item
- delete all for an action

ethitter:
The first two items were addressed by #36.

ethitter:
The list command should also provide a filter to limit by action, timestamp, or schedule.

ethitter:
We also need commands to check/reset locks

ethitter:
There should also be a column that indicates if the job is an Internal Job. It may be helpful to know which jobs will run regardless of queue size.

ethitter:
If I add a `run` command, we can block the regular cron functions.

PR #51, opened 2016-12-08
Indicate if job is an Internal Event

ethitter:
While the `a8c` prefix is currently a giveaway, that may not always be the case.

See #28


ethitter:
For the record, I have no idea why I created a branch _just_ for this.

Issue #84, opened 2017-02-08
Lock: use DB?


ethitter:

WPprodigy:
Depending on how https://github.com/Automattic/Cron-Control/pull/126#issuecomment-968522331 works out, this may be the correct route.

WPprodigy:
Closing this as it's a bit stale. Tracking in https://github.com/Automattic/Cron-Control/issues/233.

Issue #52, opened 2016-12-08
Internal Events: Introduce job to periodically scan for duplicate events


ethitter:
Should the plugin periodically scan for pending events that are non-recurring and have the same action and arguments? Could these events ever _not_ be duplicates?

ethitter:
Hopefully can be closed after #89 is merged as it won't be necessary.

See also #49.

ethitter:
Closing as #89 largely eliminated this problem, and #49 can be addressed if necessary.

Issue #49, opened 2016-12-07
Events: Detect duplicate events created with timestamps within two or three seconds


ethitter:
When the plugin is activated under significant load, it's still possible for duplicate events to be created, despite what #12 and #19 do to address this. #19 has been effective for almost all cases, but there are still situations where a duplicate is created with a timestamp that is just a few seconds off, which isn't accounted for.

To avoid any further delay when rebuilding the cron array, either the job runner, or a new Internal Job, should be created to check events for "adjacent" timestamps.

ethitter:
See also #52.

ethitter:
#89 has largely proven this unnecessary. Most, if not all, duplication is now due to improper use of Core's cron functions (not calling `wp_next_scheduled()`, for example), rather than Cron Control's deficiencies. After #114 is in use for a period, we can revisit the need for this.

WPprodigy:
Tracking in https://github.com/Automattic/Cron-Control/issues/234

Issue #37, opened 2016-12-06
Events Store: Split cache into smaller chunks when there are many events


ethitter:
If the re-rendered `cron` option is too large to store in a single cache key, split it.

PR #89, opened 2017-02-17
Convert to custom table

### LINKED ISSUE #24: CPT: Post-cache cleaning is either broken or insufficient
The output of `wp post list --post_type=a8c_cron_ctrl_event --post_status=inherit --order=asc` is unreliable as compared to that of `wp cron event list`, due to Core's caching. This doesn't impact the plugin, as it handles caching itself (#12), but the plugin does make efforts to fix cache inconsistencies, which are clearly failing. The plugin should probably adopt its own version of `clean_post_cache()`.

### LINKED ISSUE #39: CPT: Class's `convert_option()` method should have a sane upper limit
Right now, `convert_option()` will run indefinitely if asked to. Like the `get_option()` method, which has a cap, so should `convert_option()`.

### LINKED ISSUE #63: Internal Events: Should use helper functions to update CPT entries
The plugin's Internal Events currently use Core's cron functions, but should instead interact with the CPT directly, to improve performance and reduce race conditions.

### LINKED ISSUE #64: Events: Introduce functions to get an event ID from various data
There are now several places where the plugin needs to determine a post ID from some amount of event data. The plugin should have a standard set of functions to do so before it ends up with similar, redundant implementations.

### LINKED ISSUE #85: CPT: Replace with custom table that includes unique key(s)
The combination of timestamp, action, and arguments can be used to prevent duplicates at the MySQL level, rather than relying on caches for locking.

### LINKED ISSUE #90: Events Store: `_set_cron_array( array() )` won't reset store
Calling `_set_cron_array( array() )` does nothing. This appears to be a problem in the unscheduling logic.

Core uses this approach to reset Cron during PHPUnit tests, which revealed this problem. This may be a unique-enough situation to close this as wontfix.

### LINKED ISSUE #91: Events Store: consider using `schedule_event` filter to block duplicates
Returning `false`, using the `job_exists()` method, would short-circuit event creation.

See https://developer.wordpress.org/reference/hooks/schedule_event/.

### LINKED ISSUE #94: Events Store: cutoff in `get_option()` filter can send plugin into endless creation loop
In the callback that intercepts `get_option( 'cron' )` requests, there's a hard limit of 1,500 items, to control runaway cron. Unfortunately, this exacerbates runaway cron because subsequent calls to `wp_next_scheduled()` etc will fail to locate the event, so it's scheduled again, ensuring the site stays over the limit.

ethitter:
By introducing a custom table with columns and indices specific to the plugin's needs, issues with duplicate events, race conditions that create excessive events, and stale caches are all alleviated. Additionally, by adding a number of utility functions, the plugin's internal consistency is greatly improved.

Some of the changes herein:
* table is created, version stored in an option, mirroring Core's behaviour
* table includes a unique constraint to prevent duplicate entries from being added
* CLI to purge legacy CPT data is introduced
* various classes no longer contain methods to retrieve events thanks to the introduction of flexible helper functions

Outstanding concerns:
* table lacks sufficient indices; currently, they're mostly used to block duplicates, not for performance
* table is created on first load, which could lead to a race condition of table creation; I'm not sure it's safe to restrict to `is_admin()`, though, as WP will keep trying to create entries

Fixes #24 
Fixes #39 
Fixes #63 
Fixes #64 
See #84
Fixes #85 
Fixes #90 
Fixes #91
Fixes #94


ethitter:
Optimistically, duplicates will be less of an issue, obviating #49 and #52.

ethitter:
I'd still like to address #37, but don't think it's a blocker for this PR.

ethitter:
The other concern I have is around the performance of the `job_exists()` method in `Events_Store`. Some caching might be warranted.

joshbetz:
Noting the schema here:

```sql
CREATE TABLE `{$this->get_table_name()}` (
	`ID` bigint(20) unsigned NOT NULL AUTO_INCREMENT,

	`timestamp` bigint(20) unsigned NOT NULL,
	`action` varchar(255) NOT NULL,
	`action_hashed` varchar(32) NOT NULL,
	`instance` varchar(32) NOT NULL,

	`args` longtext NOT NULL,
	`schedule` varchar(255) DEFAULT NULL,
	`interval` int unsigned DEFAULT 0,
	`status` varchar(32) NOT NULL DEFAULT 'pending',

	`created` datetime NOT NULL,
	`last_modified` datetime NOT NULL,

	PRIMARY KEY (`ID`),
	UNIQUE KEY `ts_action_instance` (`timestamp`, `action` (191), `instance`, `status`)
) ENGINE=InnoDB;
```

joshbetz:
> table is created on first load, which could lead to a race condition of table creation; I'm not sure it's safe to restrict to is_admin(), though, as WP will keep trying to create entries

How hard would it be to maintain back-compat for the custom post type and force the migration to a custom table to happen via wp-cli?

includes/wp-cli/class-events.php
```diff
@@ -211,64 +222,93 @@ private function format_events( $events ) {
...
+
+		if ( $first_timestamp < $second_timestamp ) {
+			return -1;
+		}
+
+		if ( $first_timestamp > $second_timestamp ) {
+			return 1;
+		}
+
+		// If timestamps are equal, consider action
```
joshbetz: `return strnatcasecmp($first['action'], $second['action'])`?

http://php.net/manual/en/function.strnatcasecmp.php

includes/wp-cli/class-events.php
```diff
@@ -211,64 +222,93 @@ private function format_events( $events ) {
...
+			return 0;
+		}
+
+		if ( $first_timestamp < $second_timestamp ) {
+			return -1;
+		}
+
+		if ( $first_timestamp > $second_timestamp ) {
+			return 1;
+		}
```
joshbetz: Could simplify these two conditionals to

```
if ( $first_timestamp !== $second_timestamp ) {
   return $first_timestamp - $second_timetamp;
}
```

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+	 * @param int $job_id        ID of job's record
+	 * @param bool $flush_cache  Whether or not to flush internal caches after creating/updating the event
+	 * @return bool
+	 */
+	public function mark_job_record_completed( $job_id, $flush_cache = true ) {
+		global $wpdb;
+
+		$updates = array(
+			'status'   => self::STATUS_COMPLETED,
+			'instance' => wp_rand( 1000000, 999999999 ), // Breaks unique constraint, and can be recreated from entry's remaining data
```
joshbetz: Why do we need to break the unique constraint here?

ethitter: With non-recurring events scheduled for specific times, like the ones that publish scheduled posts, if the event is cleared and rescheduled, the rescheduled event can't be marked as completed until the plugin's cleanup job runs. As a result, the rescheduled event would be fired repeatedly, for up to an hour. Since instance is a hash of other data we have, overwriting it breaks the unique without destroying data.

Events aren't deleted directly for performance reasons. That's true in the CPT version, which carried over with this PR. Completed events will be cleared less-frequently after this PR--every hour instead of every ten minutes--to take advantage of indices and reduce disruptions to InnoDB's clustering.

Looking at this now, I'm not sure that `wp_rand()` is really necessary; `mt_rand()` is probably sufficient.

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+
+		if ( isset( $args['page'] ) ) {
+			$page  = max( 0, $args['page'] - 1 );
+			$offset = $page * $args['quantity'];
+		} else {
+			$offset = 0;
+		}
+
+		// Do not sort, otherwise index isn't used
+		$jobs = $wpdb->get_results( $wpdb->prepare( "SELECT * FROM {$this->get_table_name()} WHERE status = %s LIMIT %d,%d;", $args['status'], $offset, $args['quantity'] ), 'OBJECT' );
```
ethitter: Ah, bad documentation. It doesn't use an index (right now), but adding a sort changes it from a simple where query to one needing a filesort.

ethitter:
> How hard would it be to maintain back-compat for the custom post type and force the migration to a custom table to happen via wp-cli?

It wouldn't be simple. We'd need an abstraction layer to switch various lookups from one source to the other, which would be complicated by the different data format that the custom table affords.

To help with the initial table creation, the function now locks itself using `wp_cache_add()`, so the issue is less of a concern to me than when I opened the PR (2f019c6).

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+
+		return $job;
+	}
+
+	/**
+	 * Check if a job post exists
+	 *
+	 * Uses a direct query to avoid stale caches that result in duplicate events
+	 */
+	public function job_exists( $timestamp, $action, $instance, $return_id = false ) {
```
ethitter: This is one of several vestigial patterns in this plugin, which I'm not thrilled about. I'll look at fixing this.

includes/class-events-store.php
ethitter: Forgot to note that there is an index on status, care of the unique. Adding an index on status alone doesn't help, unless this query is split into one that retrieves IDs, and another that retrieves those rows using `IN`. Since it's a simple where, I'm not sure splitting is worth it.

ethitter:
> I agree we need one or two indexes -- probably `status` and maybe `action` or `action_hashed`.

I was mistaken. 😞 

`status` and `action` are part of the unique index already in place. I added a standalone index for each of the three columns, but it proved unnecessary. Explaining each of the queries the plugin uses showed that the `ts_action_instance` index was used instead of a column-specific one, even when `action_hashed` was in the query.

includes/wp-cli/class-events.php
ethitter: :man_facepalming:

ethitter:
With the exception of the weird pattern that is `job_exists()` returning an ID on demand, I've addressed all of Josh's feedback. 

More feedback welcome!!!

And yes, I'm looking at making `job_exists()` less...strange. We're living with some poor decisions I made early on in this plugin's development.

includes/class-events-store.php
ethitter: Thanks for calling this out! It turns out that `job_exists()` isn't really needed at all, just the ID-returning version of it. :man_facepalming:

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+		$cron_array = array(
+			'version' => 2, // Core versions the cron array; without this, events will continually requeue
+		);
+
+		// Get events to re-render as the cron option
+		$page     = 1;
+		$quantity = 100;
+
+		do {
+			$jobs_posts = $this->get_jobs( array(
```
mjangda: Minor: rename to `$jobs` since they're not posts anymore :)

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+
+		return $old_value;
+	}
+
+	/**
+	 * When an entry exists, don't try to create it again
+	 */
+	public function block_creation_if_job_exists( $job ) {
+		$instance = md5( maybe_serialize( $job->args ) );
+		if ( $this->job_exists( $job->timestamp, $job->hook, $instance ) ) {
```
mjangda: Minor: move out function call from `if`.

mjangda: Alternatively, following Josh's suggestion, if the method returns a `bool`, we could just do `return $this->job_exists( ... )`

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+			$offset = 0;
+		}
+
+		// Do not sort, otherwise index isn't used
+		$jobs = $wpdb->get_results( $wpdb->prepare( "SELECT * FROM {$this->get_table_name()} WHERE status = %s LIMIT %d,%d;", $args['status'], $offset, $args['quantity'] ), 'OBJECT' );
+
+		if ( is_array( $jobs ) ) {
+			$jobs = array_map( array( $this, 'format_job' ), $jobs );
+		} else {
+			$jobs = false;
```
mjangda: Can we just return an empty array?

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+		if ( ! isset( $attrs['status'] ) || ! in_array( $attrs['status'], $allowed_status, true ) ) {
+			$attrs['status'] = self::STATUS_PENDING;
+		}
+
+		// Validate attributes provided to query for a post
+		$query              = null;
+		$query_replacements = array();
+
+		if ( isset( $attrs['ID'] ) ) {
+			$query                = "SELECT * FROM {$this->get_table_name()} WHERE ID = %d";
```
mjangda: This should probably be its own function (`get_job_by_id`)?

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+
+		// Final query preparations
+		if ( 'any' !== $attrs['status'] ) {
+			$query                .= ' AND status = %s';
+			$query_replacements[]  = $attrs['status'];
+		}
+
+		$query .= ' LIMIT 1';
+
+		$query = $wpdb->prepare( $query, $query_replacements );
```
mjangda: The `query_replacements` approach seems a bit fragile. I think we should `prepare` in context (i.e. whenever we build each SQL statement).

If anything, `$query` should be an array of prepared SQL statements which is `implode`-ed with a space.

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+
+	/**
+	 * Compare two arrays and return collapsed representation of their differences
+	 *
+	 * @param array $new New cron array
+	 * @param array $old Old cron array
+	 *
+	 * @return array
+	 */
+	private function find_cron_array_differences( $new, $old ) {
```
mjangda: Minor: technically new and old are not accurate names since we swap the order to find deletes.

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+	}
+
+	/**
+	 * Remove entries for non-recurring events that have been run
+	 */
+	public function purge_completed_events( $count_first = true ) {
+		global $wpdb;
+
+		// Skip count if already performed
+		if ( $count_first ) {
```
mjangda: What do we need the `count_first` for?

includes/class-internal-events.php
```diff
@@ -113,17 +113,30 @@ public function is_internal_event( $action ) {
 	public function force_publish_missed_schedules() {
 		global $wpdb;
 
-		$missed_posts = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM {$wpdb->posts} WHERE post_status = 'future' AND post_date <= %s LIMIT 25;", current_time( 'mysql', false ) ) );
+		$page     = 1;
+		$quantity = 25;
+
+		do {
```
mjangda: Do we really need this loop here? The missed schedule event runs every minute so the next batch of 25 would just be published in the next ping? (Not to mention that something must be seriously wrong if we have more than 25 missed schedule posts.)

includes/class-internal-events.php
```diff
@@ -132,30 +145,47 @@ public function force_publish_missed_schedules() {
 	public function confirm_scheduled_posts() {
 		global $wpdb;
 
-		$future_posts = $wpdb->get_results( $wpdb->prepare( "SELECT ID, post_date FROM {$wpdb->posts} WHERE post_status = 'future' AND post_date > %s LIMIT 25;", current_time( 'mysql', false ) ) );
+		$page     = 1;
+		$quantity = 25;
+
+		do {
```
mjangda: Same as above; the loop is probably not necessary. If we expect lots of posts, can we just bump the limit to 100 or something similar (especially since we're not fetching a lot of data)?

includes/class-internal-events.php
```diff
@@ -132,30 +145,47 @@ public function force_publish_missed_schedules() {
...
+			$query        = $wpdb->prepare( "SELECT ID, post_date FROM {$wpdb->posts} WHERE post_status = 'future' AND post_date > %s LIMIT %d,%d", current_time( 'mysql', false ), $offset, $quantity );
+			$future_posts = $wpdb->get_results( $query );
 
-		if ( ! empty( $future_posts ) ) {
-			foreach ( $future_posts as $future_post ) {
-				$future_post->ID = absint( $future_post->ID );
-				$gmt_time        = strtotime( get_gmt_from_date( $future_post->post_date ) . ' GMT' );
-				$timestamp       = wp_next_scheduled( 'publish_future_post', array( $future_post->ID ) );
+			if ( empty( $future_posts ) ) {
+				break;
```
mjangda: This early `break` breaks the pattern used in other parts of the codebase.

includes/class-events-store.php
joshbetz: Does that index get used here? I think a composite index will only get used if you're querying a prefix of the index. i.e. we could use the unique key to search (timestamp), (timestamp, action), etc., but not status by itself.

https://dev.mysql.com/doc/refman/5.7/en/multiple-column-indexes.html

> If the table has a multiple-column index, any leftmost prefix of the index can be used by the optimizer to look up rows. For example, if you have a three-column index on (col1, col2, col3), you have indexed search capabilities on (col1), (col1, col2), and (col1, col2, col3).

includes/class-events-store.php
ethitter: When it's called from CLI, the count query is already run, so I wanted to avoid running it again or blocking the delete call. Now that I've described it, it should be `$force` instead.

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+			self::STATUS_PENDING,
+			self::STATUS_RUNNING,
+			self::STATUS_COMPLETED,
+		);
+	}
+
+	/**
+	 * Create the plugin's DB table when necessary
+	 */
+	protected function prepare_tables() {
```
ethitter: Initially I did, but until the table is created, every request tries to schedule the events a site defined. The `wp_cache_add()` is meant to guard against this. Instead, or perhaps in addition, the option should be set sooner, to cut down on how many processes can create the table.

ethitter:
I want to look at #94 as part of this. Two sites created nearly 4,000 entries apiece, likely thanks to that.

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+
+		// Use Core's method of creating/updating tables
+		if ( ! function_exists( 'dbDelta' ) ) {
+			require_once ABSPATH . '/wp-admin/includes/upgrade.php';
+		}
+
+		global $wpdb;
+
+		// Define schema and create the table
+		$schema = "CREATE TABLE `{$this->get_table_name()}` (
```
ethitter: I'd _really_ rather not. I'm familiar with the FG and related P2 discussion, and the arguments made there don't apply to this table, at least as I understand them. There's no use case for joining the plugin's table against any others--the IDs in `a8c_cron_control_jobs` don't correspond to any ID in any other table. If we found ourselves in a situation where a join was necessary, I'd rather we alter the table (and add appropriate indexes) than introduce a dependency on another table.

I'm also opposed to this change because I don't want to repeat the full column name throughout the codebase, nor do I really want to introduce a constant, helper, etc. to avoid repeating `a8c_cton_control_jobs_id`.

I understand the reasoning for tables on wpcom, or that might be joined against others, but don't feel it adds any value in this plugin.

includes/class-events-store.php
ethitter: My intent was to make this method a bit like `WP_Query` in its flexibility. The use case for looking a job up by ID is limited to the plugin's CLI commands, so I can move this bit somewhere specific to that context.

includes/class-events-store.php
ethitter: It's...different, that's for sure. Again, this comes from spending too much time staring at Core. With the ID-lookup bit going away, this approach won't really be needed anymore.

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+
+		dbDelta( $schema, true );
+
+		// Confirm that the table was created, and set the option to prevent further updates
+		$table_count = count( $wpdb->get_col( "SHOW TABLES LIKE '{$this->get_table_name()}'" ) );
+
+		if ( 1 === $table_count ) {
+			update_option( self::DB_VERSION_OPTION, self::DB_VERSION, true );
+		} else {
+			delete_option( self::DB_VERSION_OPTION );
```
ethitter: Yeah, I've been unsure of what to do here, because if the table doesn't exist, cron is completely broken as long as the plugin is active. If the table isn't created, the plugin shouldn't override `get|update_option()` calls. Setting an option or cache key to disable the filters in `class_init()` might be worthwhile, as well as setting the option to the expected version as a short-circuit to future upgrades. An action, which we can hook separately from mu-plugins, could handle alerting.

includes/class-events-store.php
```diff
@@ -0,0 +1,592 @@
...
+		global $wpdb;
+
+		return $wpdb->prefix . self::TABLE_SUFFIX;
+	}
+
+	/**
+	 * Build array of valid event statuses
+	 */
+	public function get_allowed_statuses() {
+		return array(
```
ethitter: I think so. There was a time when I thought this might be dynamic depending on how/where it was called, but that hasn't proven necessary.

includes/class-internal-events.php
ethitter: I'll bump the limit to 100, but I want to keep this loop. We have a few clients who schedule weeks worth of posts, and get close to the cap of 100. I'd rather confirm everything than risk missing something. This only runs every ten minutes, so I'm not that concerned about the load it'll introduce or the time it'll take to process.

includes/class-internal-events.php
ethitter: Yeah, I have no idea...

includes/class-events-store.php
ethitter: You're right, the composite index doesn't get used. In normal circumstances, there aren't enough rows to use an index on `status`, but I think it's worth adding for safety's sake.

ethitter:
@mjangda & @joshbetz - this is ready for another look. There's an open question around stampedes with the initial table creation, but I've addressed the remaining feedback and left a few thoughts on how to deal with the table.

tests/bootstrap.php
```diff
@@ -20,6 +20,15 @@ function _manually_load_plugin() {
...
 
 	require dirname( dirname( __FILE__ ) ) . '/cron-control.php';
+
+	// Plugin loads after `wp_install()` is called, so we compensate
+	// See the `class_init()` method in Events_Store for the logic behind this
+	\Automattic\WP\Cron_Control\Events_Store::instance()->prepare_table();
+	remove_filter( 'schedule_event', '__return_false' );
+	add_filter( 'pre_option_cron', array( \Automattic\WP\Cron_Control\Events_Store::instance(), 'get_option' ) );
+	add_filter( 'pre_update_option_cron', array( \Automattic\WP\Cron_Control\Events_Store::instance(), 'update_option' ), 10, 2 );
+	add_filter( 'schedule_event', array( \Automattic\WP\Cron_Control\Events_Store::instance(), 'block_creation_if_job_exists' ) );
```
mjangda: Minor: if we add `use \Automattic\WP\Cron_Control\Events_Store` to the top of this file, we can simplify the calls to this namespace:

```
add_filter( 'pre_update_option_cron', array( Events_Store::instance(), 'update_option' ), 10, 2 );
add_filter( 'schedule_event', array( Events_Store::instance(), 'block_creation_if_job_exists' ) );
```

ethitter: Good point. I honestly don't know why I didn't use `use` anywhere in the tests or CLI classes.

PR #99, opened 2017-04-13
Ensure `DOING_CRON` is set for all REST requests

### LINKED ISSUE #97: REST API: set `DOING_CRON` for `list` endpoint
For consistency, and to accommodate conditionally loading cron-related callbacks and functionality, all endpoints should set the constant `DOING_CRON`. Otherwise, callbacks may appear empty when they are not, which runs afoul of the empty-action-skipping feature (`Events::action_has_callback_or_should_run_anyway()`).

ethitter:
If code is loaded conditionally when `DOING_CRON` is set, not setting it for all of the plugin's REST endpoints can lead to unexpected behaviour.

For example, duplicate events can be created infinitely, in combination with the skip-empty-event functionality.

Fixes #97

### DIFF
--- a/includes/class-events.php
+++ b/includes/class-events.php
@@ -41,7 +41,7 @@ class Events extends Singleton {
 		define( 'DOING_CRON', true );
 
 		// When running events, allow for long-running ones, and non-blocking trigger requests
-		if ( 'run' === $endpoint ) {
+		if ( REST_API::ENDPOINT_RUN === $endpoint ) {
 			ignore_user_abort( true );
 			set_time_limit( JOB_TIMEOUT_IN_MINUTES * MINUTE_IN_SECONDS );
 		}
--- a/includes/class-events.php
+++ b/includes/class-events.php
@@ -32,7 +32,8 @@ class Events extends Singleton {
 	 */
 	public function prepare_environment() {
 		// Limit to plugin's endpoints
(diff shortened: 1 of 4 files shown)


includes/class-events.php
```diff
@@ -31,13 +31,19 @@ protected function class_init() {
 	 * This also runs before Core has parsed the request and set the \REST_REQUEST constant
 	 */
 	public function prepare_environment() {
-		if ( ! is_rest_endpoint_request( 'run' ) ) {
+		// Limit to plugin's endpoints
+		if ( ! is_rest_endpoint_request( 'list' ) && ! is_rest_endpoint_request( 'run' ) ) {
```
mjangda: Should we just have a `get_endpoint_type()` function instead?

ethitter: Yeah, probably a good idea. This is the first situation where we've needed to check multiple endpoints, but it probably won't be the last given the inconsistent behaviour this PR resolves.

ethitter:
Get out of my head! :)

I was thinking about this PR on my drive to the airport, and realizing how little sense the current approach makes. `get_endpoint_type()` should return the string to compare against the constant, and `is_rest_endpoint_request()` should check that function against the constant as well. It's silly that I'm trying to assign "nice" names to these endpoints, it just adds unnecessary complexity.

mjangda:
🚢

PR #103, opened 2017-04-18
CLI: Allow for running all events with a given action

### LINKED ISSUE #102: CLI: provide option to run all events with a given action
`wp cron-control events run` only accepts a single ID, leaving no way to run all jobs with a given action. Like the `delete` command, `run` should accept either an ID or an action.

ethitter:
Unlike the option provided by WP-CLI, this will set the lock to block regular requests from also running the event, and can provide a better UI for running many events (akin to the delete command).

Fixes #102

### DIFF
--- a/includes/class-events.php
+++ b/includes/class-events.php
@@ -270,7 +270,7 @@ class Events extends Singleton {
 	 * @param $event array Event data
 	 */
 	private function prime_event_action_lock( $event ) {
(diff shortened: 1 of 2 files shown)


ethitter:
The delete-by-action and run-by-action methods are nearly identical, save for the wording displayed  to the end user, and the method called to complete the desired process. It shouldn't be difficult to introduce a generic helper to replace much of the duplication.

ethitter:
An option to run all pending events could be nice, too.

ethitter:
See #132

PR #108, opened 2017-05-15
Allow whitelisted events to run concurrently

### LINKED ISSUE #106: Consider allowing multiple instances of the same job to run at once
On sites that queue up many of the same type of job, only running one at a time can cause large backlogs, depending on volume.

Should we allow up to the max global concurrency, dropping the per-event lock? Or, allow tweaking the per-event rate limiter on a per-event basis. The former is more set-and-forget.

ethitter:
In some circumstances, multiple events with the same action can safely run in parallel. This is usually not the case, largely due to Core's alloptions, but sometimes an event is written in a way that we can support concurrent executions.

Fixes #106

### DIFF
--- a/includes/class-events.php
+++ b/includes/class-events.php
@@ -353,7 +353,14 @@ class Events extends Singleton {
 	 * @return bool
 	 */
 	private function reset_event_lock( $event ) {
-		return Lock::reset_lock( $this->get_lock_key_for_event_action( $event ), JOB_LOCK_EXPIRY_IN_MINUTES * \MINUTE_IN_SECONDS );
+		$lock_key = $this->get_lock_key_for_event_action( $event );
+		$expires  = JOB_LOCK_EXPIRY_IN_MINUTES * \MINUTE_IN_SECONDS;
+
+		if ( isset( $this->concurrent_action_whitelist[ $event->action ] ) ) {
+			return Lock::free_lock( $lock_key, $expires );
+		} else {
+			return Lock::reset_lock( $lock_key, $expires );
+		}
 	}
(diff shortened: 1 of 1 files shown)


ethitter:
The easiest way to test this will likely be to create a large number of test posts using `wp post generate`, scheduled for the near future. Then, the `publish_future_post` action should be whitelisted for concurrency and the `a8c_cron_control_force_publish_missed_schedules` internal job disabled. With these preparations, the plugin can be executed normally, and the backlog monitored for concurrent processing.

ethitter:
ie, to allow two concurrent publish events:

```
add_filter( 'a8c_cron_control_concurrent_event_whitelist', function( $wh ) {
	$wh['publish_future_post'] = 2;

	return $wh;
} );
```

includes/class-events.php
```diff
@@ -48,6 +53,22 @@ public function prepare_environment() {
...
+	 * Some events, however, are fine to run concurrently, and should be whitelisted for such
+	 */
+	public function populate_concurrent_action_whitelist() {
+		$concurrency_whitelist = apply_filters( 'a8c_cron_control_concurrent_event_whitelist', array() );
+
+		if ( is_array( $concurrency_whitelist ) && ! empty( $concurrency_whitelist ) ) {
+			$this->concurrent_action_whitelist = $concurrency_whitelist;
+		}
+
+		unset( $concurrency_whitelist );
```
mjangda: Minor: this unset is unnecessary.

ethitter: :man_facepalming:

PR #111, opened 2017-06-15
Split queue to separate long-running events

ethitter:
When dedicated resources are available for long-running or resource-intensive events, these should be identified separately from "regular" events. This is done by providing a whitelist that event actions can be added to, which will cause the REST API response to list those separately.


ethitter:
With the introduction of the CLI runner via #114, this isn't necessary. Multiple runners, plus the queue-size controls and concurrency whitelist provided, should be sufficient to handle long-running events.

PR #112, opened 2017-06-16
Removing unused global $wpdb variable

david-binda:
The `$wpdb` variable is not being used in \Automattic\WP\Cron_Control\CLI\Events::get_events() methods and thus is not necessary to be made accessible in there.


ethitter:
Nice catch! Missed this in #89.

PR #114, opened 2017-07-05
Support running events via CLI

ethitter:
As an alternative to gathering and running events via REST API endpoints, this provides CLI commands and a Go program to accomplish the same.

The Go runner uses goroutines and workers to asynchronously gather and execute events. Events are polled for every 60 seconds, and run as soon as a worker is available. A Sysvinit script is provided to manage the runner.


ethitter:
I need to confirm that the `site list` call won't be impacted by a low default per-page value. Support for paging through site list will probably be necessary, to support large multisites.

ethitter:
An argument for network ID might be necessary, although multinetwork is pretty rare.

ethitter:
It's currently possible for long-running jobs to tie up all available workers. I'm not sure if this will be a problem in practice, or just a theoretical one. Separate queues (#111), or heartbeats from the workers, could address this.

Resource exhaustion may be another concern with the above, beyond the poor user experience of stuck cron.

ethitter:
Since all WP interactions follow the initial `wp cron-control orchestrate get-info` call, we can add a kill switch to that return data, available via a filter, option, or something else. Perhaps a WP-CLI command to set the option, too.

ethitter:
With the exception of the issue of long-running events tying up all workers ([#](https://github.com/Automattic/Cron-Control/pull/114#issuecomment-313260154)), the issues I've noted have been addressed, so this is ready for review. I'm still pondering how to deal with long-running events.

runner/runner.go
```diff
@@ -0,0 +1,366 @@
...
+func heartbeat() {
+	if heartbeatInt == 0 {
+		logger.Println("heartbeat disabled")
+		return
+	}
+
+	interval := time.Duration(heartbeatInt) * time.Second
+	time.Sleep(interval)
+
+	for true {
```
joshbetz: Minor: `true` is superfluous here. Can just be

```
for {
    // Do stuff
}
```

runner/runner.go
```diff
@@ -0,0 +1,366 @@
...
+const runEventsBreak time.Duration = time.Second * 10
+
+func init() {
+	flag.StringVar(&wpCliPath, "cli", "/usr/local/bin/wp", "Path to WP-CLI binary")
+	flag.IntVar(&wpNetwork, "network", 0, "WordPress network ID, `0` to disable")
+	flag.StringVar(&wpPath, "wp", "/var/www/html", "Path to WordPress installation")
+	flag.IntVar(&workersGetEvents, "workers-get", 1, "Number of workers to retrieve events")
+	flag.IntVar(&workersRunEvents, "workers-run", 5, "Number of workers to run events")
+	flag.IntVar(&heartbeatInt, "heartbeat", 60, "Heartbeat interval in seconds")
+	flag.StringVar(&logDest, "log", "os.Stdout", "Log path, omit to log to Stdout")
```
joshbetz: For simplicity, we could send debug logs to stdout and error logs to stderr and let whatever is running this redirect output?

runner/runner.go
```diff
@@ -0,0 +1,366 @@
...
+
+	for event := range queue {
+		workerEvents <- event
+	}
+
+	close(workerEvents)
+}
+
+func retrieveSitesPeriodically(sites chan<- site) {
+	for true {
```
joshbetz: Can just be `for {`

ethitter: Oh nice.

Now I have to resist the urge to define `evar` to `true` and use that instead... 😄

runner/runner.go
ethitter: Since this is intended to be run as a daemon controlled through Sysvinit/systemd, I wanted consistency in how the runner's options are set. That said, the answer to your question is yes, the defaults file could handle log destinations.

ethitter:
- [ ] Improve output for configuration errors

Currently just displays usage instructions, without indicating what's wrong.

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+		subcommand = append(subcommand, fmt.Sprintf("--network=%d", wpNetwork))
+	}
+
+	wpCli := exec.Command(wpCliPath, subcommand...)
+	wpOut, err := wpCli.CombinedOutput()
+	wpOutStr := string(wpOut)
+
+	if err != nil {
+		if debug {
+			logger.Printf("%s - %s", err, wpOutStr)
```
ethitter: The error from `exec.Command` has been sufficient in my experience. WP-CLI is generally very explicit when it protests.

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+		var err error
+		*path, err = filepath.Abs(*path)
+
+		if err != nil {
+			fmt.Printf("Error: %s", err.Error())
+			os.Exit(3)
+		}
+
+		if _, err = os.Stat(*path); os.IsNotExist(err) {
+			usage()
```
ethitter: Definitely, this needs improvement. #124 is open, but really should be resolved via this PR.

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+		} else if heartbeatInt > 0 {
+			atomic.AddUint64(&eventRunErrCount, 1)
+		}
+
+		time.Sleep(runEventsBreak)
+	}
+}
+
+func runWpCliCmd(subcommand []string) (string, error) {
+	subcommand = append(subcommand, "--allow-root", "--quiet", fmt.Sprintf("--path=%s", wpPath))
```
ethitter: It's necessary anytime we request data in JSON format, as it suppresses the "informational" stuff that commands include. Without it, we often don't get a string that can be unmarshalled.

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+		logger.Printf("<heartbeat eventsSucceededSinceLast=%d eventsErroredSinceLast=%d>", successCount, errCount)
+
+		time.Sleep(interval)
+	}
+}
+
+func getSites() ([]site, error) {
+	siteInfo, err := getInstanceInfo()
+	if err != nil {
+		return make([]site, 0), err
```
ethitter: Only if the function that threw the error returns the same data type, which isn't the case here. `getInstanceInfo()` returns a `siteInfo` struct. I'll check the other cases, but I believe that I've only used `make` where a different data type is called for.

includes/wp-cli/class-orchestrate-runner.php
```diff
@@ -0,0 +1,99 @@
...
+	public function run_event( $args, $assoc_args ) {
+		if ( 0 !== \Automattic\WP\Cron_Control\Events::instance()->run_disabled() ) {
+			\WP_CLI::error( __( 'Automatic event execution is disabled', 'automattic-cron-control' ) );
+		}
+
+		$timestamp = \WP_CLI\Utils\get_flag_value( $assoc_args, 'timestamp', null );
+		$action    = \WP_CLI\Utils\get_flag_value( $assoc_args, 'action',    null );
+		$instance  = \WP_CLI\Utils\get_flag_value( $assoc_args, 'instance',  null );
+
+		if ( ! is_numeric( $timestamp ) || ! is_string( $action ) || ! is_string( $instance ) ) {
```
ethitter: Good call. Goes along with #124 and related comments here.

> makes it far easier to fix issues when you know exactly what's wrong

But that takes all the fun out of debugging! ;)

includes/wp-cli/class-orchestrate-runner.php
```diff
@@ -0,0 +1,99 @@
...
+		if ( $timestamp > $now ) {
+			\WP_CLI::error( sprintf( __( 'This event is not scheduled to run until %1$s GMT', 'automattic-cron-control' ), date( TIME_FORMAT, $timestamp ) ) );
+		}
+
+		// Run the event
+		$run = \Automattic\WP\Cron_Control\run_event( $timestamp, $action, $instance );
+
+		if ( is_wp_error( $run ) ) {
+			\WP_CLI::error( $run->get_error_message() );
+		} elseif ( is_array( $run ) && isset( $run['success'] ) && true === $run['success'] ) {
```
ethitter: Nope! `run_event()` will only return a `WP_Error` or an array.

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	wpCli := exec.Command(wpCliPath, subcommand...)
+	wpOut, err := wpCli.CombinedOutput()
+	wpOutStr := string(wpOut)
+
+	if err != nil {
+		if debug {
+			logger.Printf("%s - %s", err, wpOutStr)
+			logger.Println(fmt.Sprintf("%+v\n", subcommand))
```
ethitter: Nope

ethitter:
> For the go runner, it's not actually clear to me how we're actually splitting off into multiple workers/queues since we use the same object reference for the events/queue object. Maybe worthwhile to write up something that walks through a high-level summary of the runner?

For sure. I made liberal use of goroutines and channels, so it's a bit opaque. I'll work on something.

For now, hopefully this helps. There are two worker pools. One is persistent and processes individual cron events. The second pool is transient and exists only to collect events from sites. The second pool uses a channel to pass events to the first pool. The transient pool is spawned every sixty seconds (okay, sixty seconds after the last completion) by yet another goroutine, which ensures that site addition and retirement is respected in short order. There are other channels, which are used for internal coordination. I'll distill this into the readme.

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+
+	for event := range queue {
+		workerEvents <- event
+	}
+
+	close(workerEvents)
+}
+
+func retrieveSitesPeriodically(sites chan<- site) {
+	for {
```
joshbetz: Instead of calling `time.Sleep` between runs, we can do:

```
for range time.Tick(time.Second) {
}
```

You can convert `getEventsLoops` to a `time.Duration` and pass that to `time.Tick()`.

runner/runner.go
joshbetz: Can we return `nil, err` instead of a site array?

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+	sites := make([]site, 0)
+	sites = append(sites, site{URL: siteInfo.Siteurl})
+
+	return sites, nil
+}
+
+func getInstanceInfo() (siteInfo, error) {
+	raw, err := runWpCliCmd([]string{"cron-control", "orchestrate", "runner-only", "get-info", "--format=json"})
+	if err != nil {
+		return siteInfo{}, err
```
joshbetz: Similarly to above, I would return a nil pointer instead of an empty siteInfo object.

runner/runner.go
mjangda: Makes sense; we should add an inline comment noting that.

runner/runner.go
ethitter: Neat!

includes/wp-cli/class-orchestrate-runner.php
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		// Prepare environment
+		if ( ! defined( 'DOING_CRON' ) ) {
+			define( 'DOING_CRON', true );
+		}
+
+		$now = time();
+		if ( $timestamp > $now ) {
+			\WP_CLI::error( sprintf( __( 'This event is not scheduled to run until %1$s GMT', 'automattic-cron-control' ), date( TIME_FORMAT, $timestamp ) ) );
```
ethitter: True. I'll update the message and fix the date format. There are a few unrelated commands that likely need similar updates.

runner/runner.go
ethitter: One small side-effect, which I'm not bothered by: the "sleep" now happens at the start of the loop, so the first event retrieval is "delayed."

runner/runner.go
ethitter: It won't work in this function:

> cannot use nil as type siteInfo in return argument

runner/runner.go
ethitter: Looks like it. I could've sworn that it errored, but the `make`s can be replaced with `nil`s. Returning a struct doesn't fare as well: https://github.com/Automattic/Cron-Control/pull/114/files#r131281577.

includes/class-events.php
```diff
@@ -23,15 +26,37 @@ protected function class_init() {
...
 		add_action( 'after_setup_theme', array( $this, 'populate_concurrent_action_whitelist' ) );
 	}
 
 	/**
+	 * Set initial options that control event behaviour
+	 */
+	public function prime_options() {
+		// Prime option to disable automatic execution
+		if ( is_admin() || ( defined( 'WP_CLI' ) && \WP_CLI ) || false !== get_endpoint_type() ) {
+			add_option( self::DISABLE_RUN_OPTION, 0, null, false );
```
ethitter: Given the limited applicability, I'm inclined to remove the priming entirely, and move the `get_option()` call to `run_disabled()`. The option is only checked during CLI and REST requests, so loading it on every request is unnecessary.

runner/runner.go
```diff
@@ -0,0 +1,379 @@
...
+	Instance  string
+}
+
+var (
+	wpCliPath string
+	wpNetwork int
+	wpPath    string
+
+	workersGetEvents int
+	workersRunEvents int
```
ethitter: It's fine with me. Doesn't change the flags used, which is good at this point.

runner/runner.go
joshbetz: I guess because it's expecting an object rather than a pointer.

ethitter:
I think that `getEventsLoop` should be shorter or configurable, to reduce worker idleness and support heavier use.

ethitter:
Runner should alert earlier, exit, or both, if CLI commands aren't present. Right now, the runner doesn't handle well a situation where the plugin isn't installed/activated on the target WP instance.

PR #115, opened 2017-07-05
Introduce CLI command to run events using data available from REST commands

ethitter:
`events run` only accepts an ID, while `rest-api get-queue` does not include IDs.


ethitter:
Closing in favor of #114.

PR #116, opened 2017-07-19
Add New Relic logging

philipjohn:
Use New Relic where available to monitor API requests and log some meta data.

### DIFF
--- a/includes/class-events.php
+++ b/includes/class-events.php
@@ -302,13 +302,17 @@ class Events extends Singleton {
 	private function can_run_event( $event ) {
 		// Limit to one concurrent execution of a specific action by default
 		$limit = 1;
-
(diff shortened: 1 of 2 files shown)


ethitter:
Let's wait until D6453-code has reported some data before modifying the plugin itself.

If this does prove necessary, the approach should be to include actions in Cron Control, and hook to those from VIP Go's mu-plugins. I don't want to clutter the plugin itself with code specific to our platform.

ethitter:
Thanks to D6453-code and #118, I don't think this is necessary. If we decide otherwise, this needs to be reworked to use hooks.

simonwheatley:
I think there's some benefit to naming the transactions and adding custom parameters to clearly identify the transactions in New Relic, so I advocate for refactoring this to actions and something in our MU plugins.

PR #118, opened 2017-07-19
Catch callback exceptions to avoid deadlock

### LINKED ISSUE #105: Events: a fatal error in the callback seems to stop event update
Not clear why, but we've recently noticed that fatals in the callback will prevent the event from updating, even though the update occurs before the event runs.

### LINKED ISSUE #117: Lock: locks can deadlock if event times out or fatals
Related to #105, if an event callback fatals, or exceeds the allotted time, the locks set for it won't be cleared. Over time, this can lead to the number of available "slots" being reduced as the lock will indicate that other events are occupying slots, when they really aren't.

Perhaps the lock should hold a timestamp for each slot, instead of the lock-wide timestamp used now.

ethitter:
If an event callback throws a fatal error, or exceeds the allowed execution time, the locks associated with that event's execution won't be freed, leading to deadlock. By catching these errors, we can provide useful feedback and prevent deadlock.

Fixes #105 
Fixes #117

### DIFF
--- a/includes/class-events.php
+++ b/includes/class-events.php
@@ -268,17 +268,29 @@ class Events extends Singleton {
 		$this->update_event_record( $event );
 
 		// Run the event
-		do_action_ref_array( $event->action, $event->args );
+		try {
+			do_action_ref_array( $event->action, $event->args );
+		} catch ( \Throwable $t ) {
+			$return = array(
+				'success' => false,
+				'message' => sprintf( __( 'Exception for job with action `%1$s` and arguments `%2$s` - %3$s in %4$s on line %5$d.', 'automattic-cron-control' ), $event->action, maybe_serialize( $event->args ), $t->getMessage(), $t->getFile(), $t->getLine() ),
+			);
+		}
 
 		// Free process for the next event, unless it wasn't set to begin with
 		if ( ! $force ) {
 			$this->do_lock_cleanup( $event );
 		}
 
-		return array(
-			'success' => true,
-			'message' => sprintf( __( 'Job with action `%1$s` and arguments `%2$s` executed.', 'automattic-cron-control' ), $event->action, maybe_serialize( $event->args ) ),
-		);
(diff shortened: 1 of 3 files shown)


includes/class-events.php
```diff
@@ -268,17 +268,29 @@ public function run_event( $timestamp, $action, $instance, $force = false ) {
 		$this->update_event_record( $event );
 
 		// Run the event
-		do_action_ref_array( $event->action, $event->args );
+		try {
+			do_action_ref_array( $event->action, $event->args );
+		} catch ( \Throwable $t ) {
+			$return = array(
+				'success' => false,
+				'message' => sprintf( __( 'Exception for job with action `%1$s` and arguments `%2$s` - %3$s in %4$s on line %5$d.', 'automattic-cron-control' ), $event->action, maybe_serialize( $event->args ), $t->getMessage(), $t->getFile(), $t->getLine() ),
```
nickdaugherty: Small thought on the verbage here - the message says `Exception for job...`, when it's really a `Throwable`. My concern is that someone debugging might read that as 'the code threw an `Exception` and spend a while searching for thrown `Exceptions`, when it could have been a regular fatal.

includes/class-events.php
```diff
@@ -268,17 +268,29 @@ public function run_event( $timestamp, $action, $instance, $force = false ) {
...
 		if ( ! $force ) {
 			$this->do_lock_cleanup( $event );
 		}
 
-		return array(
-			'success' => true,
-			'message' => sprintf( __( 'Job with action `%1$s` and arguments `%2$s` executed.', 'automattic-cron-control' ), $event->action, maybe_serialize( $event->args ) ),
-		);
+		// Callback didn't trigger an exception, indicating it succeeded
+		if ( ! isset( $return ) ) {
```
mjangda: Just minor coding style feedback: this is a bit counter-intuitive and very easy to miss that this is actually the success case.

I would move the lock cleanup into the `try` and return early in the catch to avoid this.

PR #123, opened 2017-07-27
Allow queue size and concurrency limits to be overridden

### LINKED ISSUE #41: Events: Consider making queue size variable
Most sites will be fine with a fixed queue size, but some may need the occasional, or ongoing, boost from the current default of 10.

### LINKED ISSUE #122: Events: make global concurrency, and perhaps other constants, configurable values
The plugin is currently rather rigid in some regards, which was meant to control resource usage. For example, overall event concurrency is set via a constant. Now that we have some real experience with the plugin, some of those constants should be configurable, within reason. For safety, constants should probably still be used.

ethitter:
Fixed sizes faired well to start, but it's now proven useful to be able to increase these limits in a controlled fashion.

Fixes #41
Fixes #122

### DIFF
--- a/includes/constants.php
+++ b/includes/constants.php
@@ -5,12 +5,18 @@ namespace Automattic\WP\Cron_Control;
 /**
  * Adjustable queue size and concurrency limits, to facilitate scaling
  */
-$job_queue_size = defined( 'CRON_CONTROL_JOB_QUEUE_SIZE' ) && is_numeric( \CRON_CONTROL_JOB_QUEUE_SIZE ) ? absint( \CRON_CONTROL_JOB_QUEUE_SIZE ) : 10;
+$job_queue_size = 10;
+if ( defined( 'CRON_CONTROL_JOB_QUEUE_SIZE' ) && is_numeric( \CRON_CONTROL_JOB_QUEUE_SIZE ) ) {
+	$job_queue_size = absint( \CRON_CONTROL_JOB_QUEUE_SIZE );
+}
 $job_queue_size = max( 1, min( $job_queue_size, 250 ) );
 define( __NAMESPACE__ . '\JOB_QUEUE_SIZE', $job_queue_size );
 unset( $job_queue_size );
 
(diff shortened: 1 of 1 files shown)


includes/constants.php
```diff
@@ -3,13 +3,22 @@
 namespace Automattic\WP\Cron_Control;
 
 /**
- * Job queues
+ * Adjustable queue size and concurrency limits, to facilitate scaling
+ */
+$job_queue_size = defined( 'CRON_CONTROL_JOB_QUEUE_SIZE' ) && is_numeric( \CRON_CONTROL_JOB_QUEUE_SIZE ) ? absint( \CRON_CONTROL_JOB_QUEUE_SIZE ) : 10;
```
mjangda: Do we need max/min protection for these constants? (e.g. what happens when someone accidentally sets the queue size to 0? Or 1000?)

includes/constants.php
```diff
@@ -3,13 +3,22 @@
...
+define( __NAMESPACE__ . '\JOB_QUEUE_SIZE', $job_queue_size );
+unset( $job_queue_size );
+
+$job_concurrency_limit = defined( 'CRON_CONTROL_JOB_CONCURRENCY_LIMIT' ) && is_numeric( \CRON_CONTROL_JOB_CONCURRENCY_LIMIT ) ? absint( \CRON_CONTROL_JOB_CONCURRENCY_LIMIT ) : 10;
```
mjangda: Minor: for readability I would avoid the ternary here and go for an explicit `if`:

```
$job_concurrency_limit = 10;
if ( defined( 'CRON_CONTROL_JOB_CONCURRENCY_LIMIT' ) && is_numeric( \CRON_CONTROL_JOB_CONCURRENCY_LIMIT ) ) {
    $job_concurrency_limit = \CRON_CONTROL_JOB_CONCURRENCY_LIMIT; // max/min from other comment would go here
}
...
```

includes/constants.php
ethitter: A floor makes sense, good catch!

I'm not sure about a ceiling, as what's an appropriate upper limit largely depends on what's being done via cron and how cron is being executed. For now, something like 250 is probably safe. We can revisit should the need arise.

PR #125, opened 2017-08-01
Chunk events queue for cacheability

### LINKED ISSUE #37: Events Store: Split cache into smaller chunks when there are many events

ethitter:
When the cron option exceeds a given size, break it into smaller slices that can be cached individually. If option size does not exceed limit, caching is unchanged.

Fixes #37

### DIFF
--- a/includes/constants.php
+++ b/includes/constants.php
@@ -37,7 +37,7 @@ const LOCK_DEFAULT_TIMEOUT_IN_MINUTES = 10;
 /**
  * Limit on size of event cache objects
  */
-$cache_bucket_size = \MB_IN_BYTES;
+$cache_bucket_size = \MB_IN_BYTES * 0.95;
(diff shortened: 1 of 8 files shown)


includes/constants.php
```diff
@@ -33,3 +33,25 @@
...
+	$cache_bucket_size = absint( \CRON_CONTROL_CACHE_BUCKET_SIZE );
+	$cache_bucket_size = max( 256 * \KB_IN_BYTES, min( $cache_bucket_size, \TB_IN_BYTES ) );
+}
+define( __NAMESPACE__ . '\CACHE_BUCKET_SIZE', $cache_bucket_size );
+unset( $cache_bucket_size );
+
+/**
+ * Limit how many buckets can be created, to avoid cache exhaustion
+ */
+$max_cache_buckets = 25;
```
ethitter: This should be lower, perhaps 5 or 6 to start.

PR #126, opened 2017-08-02
WIP: Switch locks to individual cache keys

### LINKED ISSUE #3: Lock: Add tests for `Lock` class

ethitter:
Fixes #3

### DIFF
--- a/includes/class-lock.php
+++ b/includes/class-lock.php
@@ -38,10 +38,12 @@ class Lock {
 	}
 
 	/**
+	 * Check a single-concurrency lock
 	 *
+	 * @param string $lock
+	 * @return bool
 	 */
 	private static function check_single_lock( $lock ) {
-		// Check if process can run
 		if ( self::get_lock_value( $lock ) >= 1 ) {
 			return false;
 		} else {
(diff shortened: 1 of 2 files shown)


ethitter:
The multi-concurrency locks shouldn't be freed blindly as they are in 44c722f; it's a stepping stone.

ethitter:
This was a silly approach. Instead of an array of locks, each numbered lock should be it's own cache. They expire on their own, and the class can handle some magic to make the many caches appear as a coherent lock. This has the added benefit of allowing for substitution of MySQL's `GET_LOCK()` or something Redis-backed, etc.

WPprodigy:
> Instead of an array of locks, each numbered lock should be it's own cache.

Definitely agree. While the array is better than the current situation where it's not able to really purge out abandoned locks, having all locks share a cache key leads to pretty unfortunate race conditions.

Letting locks have their own, and using `wp_cache_add()` even for further race-condition prevention, this will be much more sustainable. A "real" `wp_cache_get_multiple()` will also help a lot so it can pull them all at once.

____

For future readers, this will help solve the problem where a job is killed off (OOMs) higher up in the stack and is never able to release its lock. Since jobs will keep running and refreshing the lock timestamp, the "slot" of this OOM'd job will be forever taken until finally all slots are deadlocked and it can recognize that and wipe the all cache key out.

WPprodigy:
Closing this as it's a bit stale. Tracking in https://github.com/Automattic/Cron-Control/issues/233.

PR #128, opened 2017-08-07
Improve date internationalization

ethitter:
As noted in #114, `date()` should be replaced with `date_i18n()` anywhere that's already supporting translation.


ethitter:
Closing because as I discovered with #137, we're localizing UTC timestamps without converting them to the local time first. As a result, we show a UTC time with a localized format and timezone.

PR #130, opened 2017-08-10
Suppress errors where JSON is used

ethitter:
Orchestrate endpoint uses JSON, which displayed errors will corrupt.

### DIFF
--- a/includes/wp-cli.php
+++ b/includes/wp-cli.php
@@ -29,7 +29,7 @@ function prepare_environment() {
 
 	// Set DOING_CRON when appropriate
 	if ( isset( $cmd[1] ) && 'orchestrate' === $cmd[1] ) {
-		@ini_set( 'display_errors', false );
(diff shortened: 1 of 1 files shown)


includes/wp-cli.php
```diff
@@ -29,6 +29,7 @@ function prepare_environment() {
 
 	// Set DOING_CRON when appropriate
 	if ( isset( $cmd[1] ) && 'orchestrate' === $cmd[1] ) {
+		@ini_set( 'display_errors', '0' );
```
mjangda: Minor: but it would be useful to have an inline comment explaining why this is needed.

Maybe worth adding `@codingStandardsIgnore` as well?

ethitter: Good call: https://github.com/Automattic/Cron-Control/commit/6728765565e588674fc79d7c40fb18ecbf7fa0d0.

PR #140, opened 2017-09-04
Fix timestamp localization

### LINKED ISSUE #138: Fix uses of `date_i18n()`
Timestamps aren't localized automatically by `date_i18n()`, so currently, a localized date is displayed in CLI output, but the timezone is actually UTC, despite the blog's timezone being shown. To resolve, the timezone should not be part of the `date_i18n()` call, but rather included in the enclosing string, as it will always be UTC.

Introduced as part of #114 

See also #128

ethitter:
Currently, a UTC date will be displayed with a localized timezone, which is rather misleading. By moving the timezone (always UTC) to the translated string, the date itself can be localized without a misleading timezone be shown.

Fixes #138

### DIFF
--- a/includes/class-rest-api.php
+++ b/includes/class-rest-api.php
@@ -78,7 +78,7 @@ class REST_API extends Singleton {
 			if ( 1 === $run_disabled ) {
 				$message = __( 'Automatic event execution is disabled indefinitely.', 'automattic-cron-control' );
 			} else {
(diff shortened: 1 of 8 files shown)


ethitter:
Rather than dropping the date translations, the timezone should be moved out of `date()` and into the string, since it will always be UTC.

PR #143, opened 2017-09-10
Allow REST endpoints to be blocked should runner be used exclusively

ethitter:


WPprodigy:
Makes sense, as w/o this you have to set a secret since `false` would drop some security measures.

PR #145, opened 2017-09-22
Allow additional internal events via a constant

ethitter:
It could be helpful to allow a limited number of custom "internal events" while maintaining/protecting the plugin's events.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+composer.lock
+vendor
+.DS_Store
+Thumbs.db
+wp-cli.local.yml
+node_modules/
+*.sql
+*.tar.gz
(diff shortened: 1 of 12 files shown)


ethitter:
ie:

```
define( 'CRON_CONTROL_ADDITIONAL_INTERNAL_EVENTS', array(
	array(
		'schedule' => 'hourly',
		'action'   => 'wpcom_vip_support_remove_user_via_cron',
		'callback' => array( 'Automattic\VIP\Support_User\User', 'do_cron_cleanup' ),
	),
) );
```

The callback must be a static method or a function, as this will usually be defined in `wp-config.php`, long before plugins are loaded.

PR #151, opened 2017-11-04
Exit without an error status when event is not found

### LINKED ISSUE #150: Don't exit with an error if event is not found when running
When multiple processes are triggering cron simultaneously, sometimes an event will not be found when it is executed, because another runner picked it up.

To prevent these situations as being treated as errors (and creating noisy alerts), we can log the condition, but exit without an error.

nickdaugherty:
This situation is ‘normal’ in multi-process runners, as one runner may
have picked up the job before another job attempts it.

When that happens, we shouldn’t treat it as an error - rather log it,
and exit normally.

Fixes #150

### DIFF
--- a/includes/wp-cli/class-orchestrate-runner.php
+++ b/includes/wp-cli/class-orchestrate-runner.php
@@ -84,7 +84,13 @@ class Orchestrate_Runner extends \WP_CLI_Command {
 		$run = \Automattic\WP\Cron_Control\run_event( $timestamp, $action, $instance );
 
 		if ( is_wp_error( $run ) ) {
-			\WP_CLI::error( $run->get_error_message() );
+			$error_data = $run->get_error_data();
+
+			if ( isset( $error_data['status'] ) && 404 === $error_data['status'] ) {
+				\WP_CLI::success( $run->get_error_message() );
+			} else {
+				\WP_CLI::error( $run->get_error_message() );
+			}
 		} elseif ( isset( $run['success'] ) && true === $run['success'] ) {
 			\WP_CLI::success( $run['message'] );
 		} else {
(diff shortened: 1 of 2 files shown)


includes/wp-cli/class-orchestrate-runner.php
```diff
@@ -84,7 +84,13 @@ public function run_event( $args, $assoc_args ) {
 		$run = \Automattic\WP\Cron_Control\run_event( $timestamp, $action, $instance );
 
 		if ( is_wp_error( $run ) ) {
-			\WP_CLI::error( $run->get_error_message() );
+			$error_data = $run->get_error_data();
+
+			if ( isset( $error_data['status'] ) && 404 === $error_data['status'] ) {
+				\WP_CLI::success( $run->get_error_message() );
```
mjangda: Should this be a warning + early return?

ethitter: A warning is more appropriate, but the early return isn't needed as nothing else should happen in this function after the message is output.

ethitter:
> This situation is ‘normal’ in multi-process runners

Indeed.

https://www.youtube.com/watch?v=c2dDkR04VXY

includes/wp-cli/class-orchestrate-runner.php
nickdaugherty: 👍 switched to warning in 381d38da09bdaa4587f8058b4f56ab5a4243c94e

nickdaugherty:
NOTE - the build failures are from new PHPCS rules around commas in function args, mostly:

```
Expected 1 space after comma in function call; 4 found
```

includes/wp-cli/class-orchestrate-runner.php
```diff
@@ -87,7 +87,9 @@ public function run_event( $args, $assoc_args ) {
 			$error_data = $run->get_error_data();
 
 			if ( isset( $error_data['status'] ) && 404 === $error_data['status'] ) {
-				\WP_CLI::success( $run->get_error_message() );
+				\WP_CLI::warning( $run->get_error_message() );
+
+				exit;
```
joshbetz: Should this be `return` rather than `exit`, just to be safe? I'm not sure if this function would be called anywhere else.

ethitter: The `exit` is fine. This function shouldn't be called anywhere outside of WP-CLI; there are helpers in the `Events` class for those situations.

Issue #152, opened 2017-11-14
Perform graceful shutdown of the runner on SIGINT/SIGTERM


darthhexx:
To ensure we don't leave things half done, we need to catch the signal, set a variable and then check it in our main loop for the exit condition.

ethitter:
I added the option to disable execution to deal with this situation. It's a per-site option, but if one needed to shutdown all execution, one could filter the `a8c_cron_control_disable_run` to return `1`.

darthhexx:
@ethitter what do you think of the PR #154 ? Seems to work as expected although I haven't tested it with an actual long running CLI job.

PR #154, opened 2017-11-16
Add Cron Runner graceful service exit

darthhexx:
* Move away from ranging over timer ticks to a `waitForEpoch` utility function, which allows us to keep an eye on signal handling.
* Add global variables to monitor service routines and worker running states for graceful exit.

### DIFF
--- a/runner/init.sh
+++ b/runner/init.sh
@@ -48,7 +48,7 @@ case $1 in
   if [ -e $PIDFILE ]; then
    status_of_proc -p $PIDFILE $DAEMON "Stoppping the $NAME process" && status="0" || status="$?"
    if [ "$status" = 0 ]; then
-    start-stop-daemon --stop --quiet --oknodo --pidfile $PIDFILE
+    start-stop-daemon --stop --retry=TERM/60/KILL/5 --quiet --oknodo --pidfile $PIDFILE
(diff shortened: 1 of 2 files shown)


darthhexx:
Fixes #152

darthhexx:
Okay. It will never exit unless it feels it is time to do so :) (bar SIGKILL of course)

I'll do some more testing, because TBH I just squeezed this development in between other things and I had a feeling this may be the case.

darthhexx:
@ethitter this is fixed now, according to my testing without events anyway. Can you double check for me please?

ethitter:
@darthhexx This is better, but I'm still able to get the thing stuck while shutting down. It looks like the site retrievers don't always die, and it gets stuck waiting for them to.

darthhexx:
I have been unable to recreate the issue of site retrievers getting stuck after 36 attempts with different settings. Even with 50 of both worker types running and a get events interval of 2s they are all shutting down as expected.

Those results, together with the fact that the init script will `SIGKILL` the app after 60 seconds anyway, I am going to go ahead with this code in production.

If you have a particular set of settings and/or steps when you encounter the issue, please could you send them to me?

darthhexx:
@mjangda this is already in production, so I'd appreciate a quick review so I can merge.

PR #155, opened 2017-11-16
Initial pass at enabling alternative storage for Event_Store using hooks

sc0ttkclark:
Added a number of filters that allow for overriding `$wpdb` interaction with the Cron Control table. This would allow for many other stores to be utilized instead, using a few hooks.

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -174,6 +174,19 @@ class Events_Store extends Singleton {
 	 * Create the plugin's DB table when necessary
 	 */
 	protected function _prepare_table() {
+		/**
+		 * Allow plugins to bypass the prepare_table() process with their own.
+		 *
+		 * @param boolean $bypass
+		 *
+		 * @since TBD
+		 */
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
More work to be done since $wpdb is talking to the custom table outside of Event_Store (which the code really shouldn't be, should it?)

ethitter:
Interesting! At first glance, I like this idea.

> which the code really shouldn't be, should it?

Heh, no, it should not. The custom table was added in version 2, so I'm not entirely surprised that there are table references outside of the `Events_Store` class, but still, that's not intentional.

sc0ttkclark:
Thanks :)

Will see what I can do to adjust things everywhere else too.

mjangda:
Out of curiosity, what alternate store would you want to use?

Also, might be cleaner just to override the `Events_Store` completely and allow swapping in an alternate class (instead of a new filter for each data operation we need to do)?

WPprodigy:
Think we can close this as it's a bit stale now, but I do agree that ideally `Events_Store` itself would be the overridable piece (or perhaps a subset of it that implements the basic CRUD operations).

PR #158, opened 2018-02-01
Add CLI command and function to delete events by schedule

nickdaugherty:
Sometimes it's useful to remove all events that have a given schedule. 

Specifically, we want to use this to clean out all one-time events when syncing a database between a production and pre-production environment on VIP Go.

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -861,7 +861,7 @@ class Events_Store extends Singleton {
 	/**
 	 * Remove entries by schedule
 	 *
-	 * @param bool $count_first Should events be counted before they're deleted.
+	 * @param string $schedule Should events be counted before they're deleted.
 	 */
 	public function delete_events_by_schedule( $schedule ) {
 		global $wpdb;
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -857,6 +857,41 @@ class Events_Store extends Singleton {
 
 		return (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(ID) FROM {$this->get_table_name()} WHERE status = %s", $status ) ); // Cannot prepare table name. @codingStandardsIgnoreLine
 	}
+
+	/**
+	 * Remove entries for non-recurring events that have been run
+	 *
+	 * @param bool $count_first Should events be counted before they're deleted.
+	 */
+	public function delete_events_by_schedule( $schedule ) {
+		global $wpdb;
+
+		return $wpdb->delete(
+			$this->get_table_name(), array(
+				'schedule' => $schedule,
+			)
+		);
+	}
+
+	/**
+	 * Count number of events with a given schedule
+	 *
+	 * @param string $schedule Event schedule to count.
+	 * @return int|false
+	 */
+	public function count_events_by_schedule( $schedule ) {
+		global $wpdb;
+
+		$sql = '';
+
+		if ( is_null( $schedule ) ) {
+			$sql = "SELECT COUNT(ID) FROM {$this->get_table_name()} WHERE schedule IS NULL"; // Cannot prepare table name. @codingStandardsIgnoreLine
+		} else {
+			$sql = $wpdb->prepare( "SELECT COUNT(ID) FROM {$this->get_table_name()} WHERE schedule = %s", $schedule ); // Cannot prepare table name. @codingStandardsIgnoreLine
+		}
+
+		return (int) $wpdb->get_var( $sql );
+	}
 }
 
 Events_Store::instance();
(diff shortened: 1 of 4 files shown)


ethitter:
Two passing thoughts:

* dealing with #102 and inverting that work to handle deletes may be more-widely applicable, as I can't recall needing to delete by schedule in any other scenario, but deleting by action has come up before;
* this could use an internal-cache clear after the delete runs, which should probably be a separate issue as existing commands need this change too.

includes/wp-cli/class-events.php
```diff
@@ -634,6 +639,27 @@ private function delete_completed_events( $args, $assoc_args ) {
...
 	}
+
+	/**
+	 * Delete events that have a particular schedule
+	 *
+	 * @param array $args Array of positional arguments.
+	 * @param array $assoc_args Array of flags.
+	 */
+	private function delete_events_by_schedule( $args, $assoc_args ) {
+		if ( 'null' === strtolower( $assoc_args['schedule'] ) ) {
```
mjangda: I guess the `null` value is for events that are not recurring? Might be a good idea to have some inline docs making that clear

includes/wp-cli/class-events.php
```diff
@@ -634,6 +639,27 @@ private function delete_completed_events( $args, $assoc_args ) {
...
+			$assoc_args['schedule'] = null;
+		}
+
+		$count = \Automattic\WP\Cron_Control\count_events_by_schedule( $assoc_args['schedule'] );
+
+		/* translators: 1: Event count */
+		\WP_CLI::confirm( sprintf( _n( 'Found %s single event to remove. Continue?', 'Found %s single events to remove. Continue?', $count, 'automattic-cron-control' ), number_format_i18n( $count ) ) );
```
mjangda: The use of `single event` is somewhat confusing here.

includes/class-events-store.php
```diff
@@ -857,6 +857,39 @@ public function count_events_by_status( $status ) {
...
+			$this->get_table_name(), array(
+				'schedule' => $schedule,
+			)
+		);
+	}
+
+	/**
+	 * Count number of events with a given schedule
+	 *
+	 * @param string $schedule Event schedule to count.
```
mjangda: The param can also be `null`.

includes/functions.php
```diff
@@ -146,6 +146,16 @@ function count_events_by_status( $status ) {
 	return Events_Store::instance()->count_events_by_status( $status );
 }
 
+/**
+ * Count events with a given schedule
+ *
+ * @param string $schedule Schedule to count.
```
mjangda: The param can also be `null`

includes/class-events-store.php
```diff
@@ -857,6 +857,39 @@ public function count_events_by_status( $status ) {
 
 		return (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(ID) FROM {$this->get_table_name()} WHERE status = %s", $status ) ); // Cannot prepare table name. @codingStandardsIgnoreLine
 	}
+
+	/**
+	 * Remove entries by schedule
+	 *
+	 * @param string $schedule Should events be counted before they're deleted.
```
mjangda: Incorrect inline docs?

includes/wp-cli/class-events.php
nickdaugherty: Ya - also open to other terms for the name...`single`, `now`, `none`?

includes/class-events-store.php
nickdaugherty: 4b482d4

includes/functions.php
nickdaugherty: 4b482d4

includes/class-events-store.php
nickdaugherty: Oops, fixed in 4b482d4

includes/wp-cli/class-events.php
nickdaugherty: 9ab6261

GaryJones:
@nickdaugherty Unit test failing:

> 1) Automattic\WP\Cron_Control\Tests\Events_Store_Tests::test_delete_events_by_schedule
> Unexpected incorrect usage notice for wpdb::prepare
> Failed asserting that an array is empty.

WPprodigy:
> Specifically, we want to use this to clean out all one-time events when syncing a database between a production and pre-production environment on VIP Go.

This is a good goal. I think maybe we don't need it to be a top-level CLI though. W/ the incoming changes (and a few adjustments), we could do the cleanup inside our existing "data sync cleanup" cli w/ logic that looks something like this:

```
$events = Events::query( [ 'schedule' => false ] );
foreach ( $events as $event ) {
  $event->delete();
}

// Or a more performant (single SQL statement), but less flexible, api:
$events = Events::instance()->delete_non_recurring_events();
```

PR #162, opened 2018-03-27
Update runner documentation

discoinfiltrator:
A few updates that would have helped me as I was implementing this.


WPprodigy:
Sorry this was never merged! The runner is moving to https://github.com/Automattic/cron-control-runner, and I believe it will just be the CLI, with any daemon-related pieces being custom implementation choices. Other options include starting the process in the background, like:

```
nohup /usr/local/bin/cron-control-runner -wp-path=/wp -wp-cli-path=/usr/local/bin/wp -get-sites-interval=120s -get-events-interval=60s &>/tmp/cron-runner.log &
```

PR #164, opened 2018-04-09
Only run get_jobs query once

### LINKED ISSUE #163: Fix get_option() for tons of jobs
The public `Events_Store::get_option()` function gets exponentially slower as the jobs table grows. You can see in the example below that high offsets with low limits still look at all the rows.  The number of queries grows linearly with the number of pending jobs and the number of rows returned per query also grows linearly.

```
mysql> explain SELECT * FROM wp_a8c_cron_control_jobs WHERE status = 'pending' LIMIT 1600,100;
+------+-------------+--------------------------+------+---------------+--------+---------+-------+------+-----------------------+
| id   | select_type | table                    | type | possible_keys | key    | key_len | ref   | rows | Extra                 |
+------+-------------+--------------------------+------+---------------+--------+---------+-------+------+-----------------------+
|    1 | SIMPLE      | wp_a8c_cron_control_jobs | ref  | status        | status | 34      | const | 5413 | Using index condition |
+------+-------------+--------------------------+------+---------------+--------+---------+-------+------+-----------------------+
1 row in set (0.04 sec)
```

I think we could make an incremental improvement here by querying all the rows in 1 query with no limit. Better would be a high, but reasonable limit so we don't have to worry about ever returning 10000 rows, for example.

We could likely do even better by doing some math with the primary key. I'm thinking something like:

```
$id = SELECT ID FROM wp_a8c_cron_control_jobs WHERE `status` = 'pending' LIMIT 1;
$jobs = SELECT * FROM wp_a8c_cron_control_jobs WHERE `status` = 'pending' AND ID > $id AND ID < $id + 200 LIMIT 100; // Pad filter by 2x because we don't know how many complete jobs are in here
$next_id = $jobs[ length( $jobs ) - 1 ][ 'ID' ];
```

joshbetz:
When there are thousands or tens of thousands of pending jobs, each call
to `get_option( 'cron' )` ends up calling `get_jobs()` tens or even
hundreds of times. The limited queries aren't so much faster that
calling it multiple times is more performant than one big query.
One reasonable query would be best so we're going to try one query with
LIMIT 1000 first.

Fixes #163

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -264,50 +264,39 @@ class Events_Store extends Singleton {
 		);
 
 		// Get events to re-render as the cron option.
-		$page     = 1;
-		$quantity = 100;
+		$jobs = $this->get_jobs(
+			array(
+				'status'   => self::STATUS_PENDING,
+				'quantity' => 1000, // Only first 1000 jobs or this gets slow.
+			)
+		);
 
-		do {
-			$jobs = $this->get_jobs(
-				array(
-					'status'   => self::STATUS_PENDING,
-					'quantity' => $quantity,
-					'page'     => $page++,
-				)
+		// Loop through results and built output Core expects.
+		foreach ( $jobs as $job ) {
+			// Alias event timestamp.
+			$timestamp = $job->timestamp;
+
+			// If timestamp is invalid, event is removed to let its source fix it.
+			if ( $timestamp <= 0 ) {
+				$this->mark_job_record_completed( $job->ID );
+				continue;
+			}
+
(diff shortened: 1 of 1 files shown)


ethitter:
This is probably going to break on at least one site, because if any site has more than 1,000 pending events and this excludes them, Core will continually reschedule the events it believes are missing. That's why all events are retrieved every time, and the cache is chunked if needed.

5.0 may introduce some new hooks that won't require rebuilding the cron option in its entirety every time, but until then, we've already had several issues related to events being considered missing, which led to runaway event creation.

PR #165, opened 2018-04-09
Reduce number of queries from get_option

### LINKED ISSUE #163: Fix get_option() for tons of jobs

joshbetz:
By reducing the number of queries by 50x, we will reduce the time spent in this function and increase performance. I'm going with 5000 because we consider it an upper limit for the number of jobs you can have. If sites have more than 5000 pending events, cron gets disabled and there are some alerts.

Fixes #163

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -34,6 +34,8 @@ class Events_Store extends Singleton {
 	);
 
 	const CACHE_KEY = 'a8c_cron_ctrl_option';
+	
+	private $invalid_cache = false;
 
 	/**
 	 * Whether or not event creation is temporarily blocked
@@ -257,11 +259,12 @@ class Events_Store extends Singleton {
 		// remotely multiple times per request (even from the
 		// object cache).
 		static $cron_array;
-		if ( $cron_array ) {
+		if ( $cron_array && ! $this->invalid_cache ) {
 			return $cron_array;
 		}
 		
 		// Use cached value when available.
+		$this->invalid_cache = false;
 		$cached_option = $this->get_cached_option();
 
 		if ( false !== $cached_option ) {
@@ -800,6 +803,7 @@ class Events_Store extends Singleton {
 	 * Delete the cached representation of the cron option
 	 */
 	public function flush_internal_caches() {
+		$this->invalid_cache = true;
 		return wp_cache_delete( self::CACHE_KEY );
 	}
 
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -265,7 +265,7 @@ class Events_Store extends Singleton {
 
 		// Get events to re-render as the cron option.
 		$page     = 1;
-		$quantity = 100;
+		$quantity = 5000;
 
 		do {
 			$jobs = $this->get_jobs(
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -40,7 +40,7 @@ class Events_Store extends Singleton {
 	 *
 	 * @var bool
 	 */
-	private $invalid_cache = false;
+	private $is_option_cache_valid = false;
 
 	/**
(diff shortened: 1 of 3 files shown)


joshbetz:
In a test of 5000 `wp_version_check` jobs, this reduced the number of queries on a wp-admin page from 691 to 6 and reduced the page generation time from 2s to 0.002s.

joshbetz:
This introduces a bug where `get_option()`, `update_option()`, `get_option()` will return the wrong results. I added the static variable cache because I was still seeing ~10 of the same query on every wp-admin page, which seems like a bug.

joshbetz:
Noting the same test is failing in other PRs. I think the tests are just broken.

joshbetz:
It might be easier to read if we rename the `invalid_cache` variable to `valid_option_cache` and reverse the logic.

includes/class-events-store.php
```diff
@@ -251,7 +258,18 @@ public function remove_multisite_table( $tables_to_drop ) {
 	 * Override cron option requests with data from custom table
 	 */
 	public function get_option() {
+
+		// If this thread has already generated the cron array,
+		// use the copy from local memory. Don't fetch this list
+		// remotely multiple times per request (even from the
+		// object cache).
+		static $cron_array;
```
mjangda: Nice improvement.

includes/class-events-store.php
```diff
@@ -251,7 +258,18 @@ public function remove_multisite_table( $tables_to_drop ) {
...
+		if ( $cron_array && ! $this->invalid_cache ) {
+			return $cron_array;
+		}
+
 		// Use cached value when available.
+		$this->invalid_cache = false;
```
mjangda: Could rename this to `$is_cache_valid` which would make it's purpose and use a bit clearer.

mjangda:
> Noting the remaining failure is also failing in other PRs. I think that part of the test is just broken.

Looks related to the PHPCS warnings?

joshbetz:
Fixed PHPCS issues

joshbetz:
```
wp> count( get_option( 'cron' ) );
int(1261)
wp> wp_schedule_single_event( time() + 60, 'wp_version_check' );
NULL
wp> count( get_option( 'cron' ) );
int(1262)
```

PR #166, opened 2018-04-11
Preserve keys when slicing the array

david-binda:
In PHP, the array union operator (+) ignores even numerical keys from the second array if those are already present in the left hand one. Thus in case we are slicing an array w/o preserving keys, we are then trying to merge two arrays with identical numerical keys. And thus, no union is happening.

This commit is fixing the issue by preserving the keys. Another apporach would be to use `array_merge` which is appending identical numberic keys instead of ignoring those, but preserving the keys feels like a better approach, when we are dealing with cached and sliced array.

This should make #165 obsolet

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -783,7 +783,7 @@ class Events_Store extends Singleton {
 
 		for ( $i = 1; $i <= $buckets; $i++ ) {
 			$offset    = ( $i - 1 ) * $segment_size;
-			$slice     = array_slice( $option_flat, $offset, $segment_size );
+			$slice     = array_slice( $option_flat, $offset, $segment_size, true );
 			$cache_key = $this->get_cache_key_for_slice( $incrementer, $i );
 
 			wp_cache_set( $cache_key, $slice, null, 1 * \HOUR_IN_SECONDS );


joshbetz:
Do the cache keys matter? It doesn't look like we use them when we put the array back together.

https://github.com/Automattic/Cron-Control/pull/166/files#diff-5ce18b52430f3c75a9d5bfa7deb87b3bR739

It looks like the array slicing is not guaranteed to work. For example, if 1 array entry is over 1MB, it will still not be cached here.

david-binda:
> Do the cache keys matter? It doesn't look like we use them when we put the array back together.

Yes, they do. I must have not explained the issue properly. From http://php.net/manual/en/language.operators.array.php

> The + operator returns the right-hand array appended to the left-hand array; for keys that exist in both arrays, the elements from the left-hand array will be used, and the matching elements from the right-hand array will be ignored.

The "ignored" here applies even to numerical keys. Vs. `array_merge` (via http://php.net/manual/en/function.array-merge.php )

> Merges the elements of one or more arrays together so that the values of one are appended to the end of the previous one. It returns the resulting array.

> If the input arrays have the same string keys, then the later value for that key will overwrite the previous one. If, however, the arrays contain numeric keys, the later value will not overwrite the original value, but will be appended.

> It looks like the array slicing is not guaranteed to work. For example, if 1 array entry is over 1MB, it will still not be cached here.

How so? This bit seems to be securing it, up to 5M, at least accordingly to my reading:

https://github.com/Automattic/Cron-Control/blob/1b21862229cdc4c6ce21881dc62be95042bb563a/includes/class-events-store.php#L760,L761

https://github.com/Automattic/Cron-Control/blob/1b21862229cdc4c6ce21881dc62be95042bb563a/includes/class-events-store.php#L777,L782

https://github.com/Automattic/Cron-Control/blob/9022943f963c037987acd7923efb3755a5b3058d/includes/constants.php#L42,L51

WPprodigy:
Oh, this is actually quite big. The keys in `$option_flat` are integers, but like the docs say they are reset when slicing w/o this last param.

The cache buckets are pieced together w/ the aforementioned `newarray = array1 + array2` type concat: https://github.com/Automattic/Cron-Control/blob/1b21862229cdc4c6ce21881dc62be95042bb563a/includes/class-events-store.php#L739

So this is what we want (w/ patch):

```
$array = ['zero', 'one', 'two', 'three', 'four', 'five', 'six'];

wp> $array1 = array_slice($array, 4, 3, true );
=> array(3) {
  [4]=>
  string(4) "four"
  [5]=>
  string(4) "five"
  [6]=>
  string(3) "six"
}

wp> $array2 = array_slice($array, 0, 4, true );
=> array(4) {
  [0]=>
  string(4) "zero"
  [1]=>
  string(3) "one"
  [2]=>
  string(3) "two"
  [3]=>
  string(5) "three"
}

$new_array = [];
$new_array +=  $array1;
$new_array +=  $array2;

wp> $new_array
=> array(7) {
  [4]=>
  string(4) "four"
  [5]=>
  string(4) "five"
  [6]=>
  string(3) "six"
  [0]=>
  string(4) "zero"
  [1]=>
  string(3) "one"
  [2]=>
  string(3) "two"
  [3]=>
  string(5) "three"
}
```

Whereas right now, this is what we get:

```
$array1 = array_slice($array, 4, 3 );
$array2 = array_slice($array, 0, 4 );

$new_array +=  $array1;
$new_array +=  $array2;

wp> $new_array;
=> array(4) {
  [0]=>
  string(4) "four"
  [1]=>
  string(4) "five"
  [2]=>
  string(3) "six"
  [3]=>
  string(5) "three"
}
```

😬 

Furthermore, this means are attempt to get from cache pretty much always fails when there is more than 1 bucket: https://github.com/Automattic/Cron-Control/blob/1b21862229cdc4c6ce21881dc62be95042bb563a/includes/class-events-store.php#L742-L745

WPprodigy:
I'm quite sure this is at least one of the major causes of "runaway events" now, where an event just keeps registering itself over and over hundreds of times, b/c WP was telling the app it didn't exist.

WPprodigy (approved):

PR #168, opened 2018-06-14
 Ignore the "no resource" errors for sitemap generation

darthhexx:
We do not want to count these as errors and have error stats generated, so we simply ignore any errors of this ilk since they are expected to hit the limit and be done in batches.


WPprodigy:
Will be taking a closer look at this in https://github.com/Automattic/cron-control-runner/issues/2

PR #174, opened 2019-01-13
Support WP 5.1's new cron hooks

### LINKED ISSUE #173: Events: Use preflight filters added in WP 5.1
**Task**
When setting and getting Cron jobs, make use of the new WP filters: `pre_schedule_event`, `pre_reschedule_event`, `pre_unschedule_event`, `pre_clear_scheduled_hook`, `pre_unschedule_hook`, `pre_get_scheduled_event`, `pre_next_scheduled` and `pre_get_ready_cron_jobs`.

**Background**
In WordPress 5.1, new filters will be added to preflight event scheduling to allow for custom Cron storage systems.

The filters were added in [WordPress#32656](https://core.trac.wordpress.org/ticket/32656) (see WordPress/wordpress-develop@a32ea2d35d2fcf7bccf54ff7abfad9a2b49ee360) and [WordPress#45797](https://core.trac.wordpress.org/ticket/45797) (see WordPress/wordpress-develop@0bc20594c78369a6778635cd58f2b90c57428d6d).

ethitter:
5.1 will introduce a new set of filters that we can use to intercept all cron-scheduling operations. 

Outstanding items:
- [ ] Introduce helper to replace `get_option( 'cron' )` calls with `Events_Store` methods.
- [ ] Tests

Fixes #173


includes/class-events-store-cron-filters.php
```diff
@@ -0,0 +1,264 @@
...
+	 * Register hooks to intercept events before storage.
+	 */
+	protected function class_init() {
+		add_filter( 'pre_schedule_event', [ $this, 'filter_event_scheduling' ], 10, 2 );
+		add_filter( 'pre_reschedule_event', [ $this, 'filter_event_rescheduling' ], 10, 2 );
+		add_filter( 'pre_unschedule_event', [ $this, 'filter_event_unscheduling' ], 10, 4 );
+		add_filter( 'pre_clear_scheduled_hook', [ $this, 'filter_clear_scheduled_hook' ], 10, 3 );
+		add_filter( 'pre_unschedule_hook', [ $this, 'filter_unchedule_hook' ], 10, 2 );
+		add_filter( 'pre_get_scheduled_event', [ $this, 'filter_event_retrieval' ], 10, 4 );
+		add_filter( 'pre_next_scheduled', [ $this, 'filter_next_scheduled' ], 10, 4 );
```
peterwilsoncc: This filter was removed in WordPress/wordpress-develop@554ee5b53c2c16ec4dc7255b5839205b8236dbf1 and the logic for finding the next scheduled event moved to `wp_get_scheduled_event()` which can be hijacked using the `pre_get_scheduled_event` filter.

ethitter:
I'm seeing a blocking issue while testing this against 5.1. Events are being duplicated at such a rate that I suspect the plugin's new checks for next-scheduled events is flawed.

peterwilsoncc:
@ethitter I suspect that's the result of 990f3666c67ef5039bd96e3f3079494422df76a6.

If `$timestamp === null` in `filter_event_retrieval()` the function ought to return the next instance of the a job:

```mysql
SELECT timestamp 
FROM {$table_name} 
WHERE 
  action = %s AND 
  instance = %s AND 
  status = %s 
ORDER BY timestamp ASC LIMIT 1
```

If a timestamp is passed, then the particular job is searched for.

ethitter:
@peterwilsoncc Good catch! I pushed a fix that partially restores the next-occurence method that I'd removed during the 5.1 beta rounds. 

I'm now running this on my multisite and things are looking fine. cc @mjangda

includes/class-events-store-cron-filters.php
```diff
@@ -0,0 +1,241 @@
...
+ * Class Option_Intercept
+ */
+class Events_Store_Cron_Filters extends Singleton {
+	/**
+	 * Register hooks to intercept events before storage.
+	 */
+	protected function class_init() {
+		add_filter( 'pre_schedule_event', [ $this, 'filter_event_scheduling' ], 10, 2 );
+		add_filter( 'pre_reschedule_event', [ $this, 'filter_event_rescheduling' ], 10, 2 );
+		add_filter( 'pre_unschedule_event', [ $this, 'filter_event_unscheduling' ], 10, 4 );
```
nickdaugherty: @ethitter What do you think about naming the callbacks with the filter name? So `function filter_pre_unschedule_event()` for example. That would make it easier to know when they are firing at a glance without referring back here.

peterwilsoncc:
I missed the use of `maybe_(un)?serialize` on the first pass of this.

WP Core uses `(un)?serialize` for setting the instance hash and saving the arguments in the native cron array, rather than the maybe functions. Using the maybe functions for the new filters may cause a mismatch between these and items saved via the option filter, depending on how you've been saving them already.

Core always serializes them to ensure an integer is always an integer, etc. It shouldn't make a difference as arguments should always be an array but there are some deprecations in WP Core that allow for them not to be.

includes/class-events-store-cron-filters.php
ethitter: That's fine with me. I believe that in an earlier iteration, some of the callbacks were shared between filters, but that's no longer the case.

ethitter:
> Question - in the PR you noted the TODO for "Introduce helper to replace get_option( 'cron' ) calls with Events_Store methods." - can you expand on that a bit? Happy to look into it.

I noticed that in a number of places, I relied on the plugin's filtering of the cron option to retrieve the cron array. That (somewhat) made sense when filtering `get_option( 'cron' )` was the canonical way to interact with the store, but with these new filters bypassing the cron option filtering, I thought it might be time to have a proper helper that the plugin can use in place of `get_option()` when it makes sense.

> I also noticed that the tests are still doing the `md5( maybe_serialize() )` for instance IDs - do we want to update those to use the new method?

Good catch, those should use the new method too.

> I missed the use of `maybe_(un)?serialize` on the first pass of this.

Yeah, I'm not sure why I deviated from what Core uses, and I can't think of any reason not to switch to the PHP-native functions.

ethitter:
Will continue in #178.

PR #175, opened 2019-02-28
Pass status code to wp_send_json_error()

WPprodigy:
WP 4.7 added the ability to pass the status code to `wp_send_json_error()` / `wp_send_json()`:
https://developer.wordpress.org/reference/functions/wp_send_json_error/

Seemed odd to be setting a 403 and then passing 400 with the WP_Error, so just stripping out that part.

### DIFF
--- a/includes/class-main.php
+++ b/includes/class-main.php
@@ -86,17 +86,13 @@ class Main extends Singleton {
 
 	/**
 	 * Block direct cron execution as early as possible
+	 *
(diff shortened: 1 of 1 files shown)


GaryJones:
Has a merge conflict that will need fixing.

WPprodigy:
The overall goal here is to help avoid some errors about headers already being sent under some situations. This should do that for us since wp_send_json() checks first before setting headers.

GaryJones (approved):
✅

PR #176, opened 2019-03-07
Enable the running of remote wp cli commands

darthhexx:
* Listen for incoming request for remote WP CLI commands on port 22122 if a token is provided.
* Requires a hexadecimal GUID in the handshake packet.
* Blacklisting of "dangerous" core commands.
* Logs all data between caller and WP CLI to disk for an audit trail.
* Supports user terminal size and resizing.
* Supports JSON or Text file or stdout logging.
* Send the WP-CLI pseudo command "vip-go-retrieve-remote-logs" to stream the logs of a completed WP-CLI process.
* If the same Guid connects and the WP-CLI is still running, then reattach and send the catchup data and continue to stream new data.
* Disconnects/reconnects can happen as many times as required.
* Multiple TCP clients can connect to the same running WP-CLI process simultaneously.
* Application will wait for up to 90 seconds for a WP-CLI command to finish before exiting.

### DIFF
--- a/runner/remote.go
+++ b/runner/remote.go
@@ -87,7 +87,6 @@ func authConn(conn *net.TCPConn) {
 		}
 
 		size := len(data)
-		logger.Print("received data:", string(data[:size]))
 
 		newlineChars := 1
 		if 1 < size && 0xd == (data[size-2 : size-1])[0] {
--- a/runner/remote.go
+++ b/runner/remote.go
@@ -15,7 +15,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/kr/pty"
+	"github.com/creack/pty"
 	"golang.org/x/crypto/ssh/terminal"
 )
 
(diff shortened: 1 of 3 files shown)


runner/remote.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+}
+
+func authConn(conn *net.TCPConn) {
+	elems := make([]string, 0)
+	var rows, cols uint64
+	var Guid string
+
+	for {
+		data := make([]byte, 256)
```
dchymko: Noting that the `256` here limits the length of the wp-cli command to around ~150 chars

darthhexx: Switched to using bufio, so the effective total packet size is now 64k.

joshbetz:
I'm looking at this between Platform requests. Code looks pretty good so far. We're doing some testing with the CLI + API + runner on a test site and there are a couple of edge cases in particular that I'm going to be looking into.

PR #177, opened 2019-04-12
Allow passing a timestamp to internal events

WPprodigy:
The documentation update explains this pretty well. The main use case is needing to setup a new internal event across a lot of sites, and wanting to stagger them a bit as to avoid both heavy load on multi-sites and on other servers for cases where we make external requests in the job.

Example:

```
define( 'CRON_CONTROL_ADDITIONAL_INTERNAL_EVENTS', array(
	array(
		'schedule'  => 'hourly',
		'action'    => 'wpcom_vip_test_cron_control_job',
		'callback'  => 'wpcom_vip_test_cron_control_job',
		'timestamp' => strtotime( sprintf( '+%d minutes', mt_rand( 1, 60 ) ) ),
	)
) );
```

This only affects the first time the job is registered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,12 +37,15 @@ In `wp-config.php` or a similarly-early and appropriate place, define `CRON_CONT
 ```
 define( 'CRON_CONTROL_ADDITIONAL_INTERNAL_EVENTS', array(
 array(
-	'schedule' => 'hourly',
-	'action'   => 'do_a_thing',
-	'callback' => '__return_true',
+	'schedule'  => 'hourly',
+	'action'    => 'do_a_thing',
+	'callback'  => '__return_true',
+	'timestamp' => strtotime( '+60 seconds' ),
 ),
 ) );
 ```
+
+The `timestamp` parameter is optional, and will default to +60 seconds if left off. Useful in cases where you would like to stagger the intial run of a new internal event across sites.
 
 Due to the early loading (to limit additions), the `action` and `callback` generally can't directly reference any Core, plugin, or theme code. Since WordPress uses actions to trigger cron, class methods can be referenced, so long as the class name is not dynamically referenced. For example:
 
--- a/includes/class-internal-events.php
+++ b/includes/class-internal-events.php
@@ -148,13 +148,12 @@ class Internal_Events extends Singleton {
 	 * Schedule internal events
 	 */
 	public function schedule_internal_events() {
-		$when = strtotime( sprintf( '+%d seconds', JOB_QUEUE_WINDOW_IN_SECONDS ) );
-
 		$schedules = wp_get_schedules();
 
 		foreach ( $this->internal_events as $event_args ) {
 			if ( ! wp_next_scheduled( $event_args['action'] ) ) {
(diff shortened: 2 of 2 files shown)


includes/class-internal-events.php
```diff
@@ -148,21 +148,20 @@ public function register_internal_events_schedules( $schedules ) {
 	 * Schedule internal events
 	 */
 	public function schedule_internal_events() {
-		$when = strtotime( sprintf( '+%d seconds', JOB_QUEUE_WINDOW_IN_SECONDS ) );
```
rogertheriault: it seems a bit more readable to retain this and maybe just change to $when_default

includes/class-internal-events.php
```diff
@@ -148,21 +148,20 @@ public function register_internal_events_schedules( $schedules ) {
...
-
 		$schedules = wp_get_schedules();
 
 		foreach ( $this->internal_events as $event_args ) {
 			if ( ! wp_next_scheduled( $event_args['action'] ) ) {
-				$interval = array_key_exists( $event_args['schedule'], $schedules ) ? $schedules[ $event_args['schedule'] ]['interval'] : 0;
+				$interval  = array_key_exists( $event_args['schedule'], $schedules ) ? $schedules[ $event_args['schedule'] ]['interval'] : 0;
+				$timestamp = isset( $event_args['timestamp'] ) ? (int) $event_args['timestamp'] : strtotime( sprintf( '+%d seconds', JOB_QUEUE_WINDOW_IN_SECONDS ) );
```
rogertheriault: It seems to me that keeping this var `$when` makes it clearer its purpose (when, when_start, start, etc.)

README.md
```diff
@@ -37,13 +37,16 @@ In `wp-config.php` or a similarly-early and appropriate place, define `CRON_CONT
...
-	'callback' => '__return_true',
+	'schedule'  => 'hourly',
+	'action'    => 'do_a_thing',
+	'callback'  => '__return_true',
+	'timestamp' => strtotime( '+60 seconds' ),
 ),
 ) );
 ```
 
+The `timestamp` parameter is optional, and will default to +60 seconds if left off. Useful in cases where you would like to stagger the intial run of a new internal event across sites.
```
rogertheriault: Should this read "will default to JOB_QUEUE_WINDOW_IN_SECONDS from now"?

rogertheriault: Also, "initial" I presume

includes/class-internal-events.php
WPprodigy: My thought process was to mirror the `wp_schedule_event()` naming: https://developer.wordpress.org/reference/functions/wp_schedule_event/. So those familiar with one could easily recognize the other.

README.md
WPprodigy: True, but since `JOB_QUEUE_WINDOW_IN_SECONDS` defaults to +60 seconds it's not lying :), and that felt more useful in the example.

WPprodigy:
Can close this out for now, and revisit if needed in the future. Internal events should be needed less and less moving forward.

PR #185, opened 2020-02-27
Travis: Set distribution and MySQL service

GaryJones:
This Travis config file was last amended when Trusty was the default Linux distribution for Travis. The default is now Xenial, and both Xenial and Bionic no longer include the MySQL service by default, and as such, the unit tests fail to initialise and all PRs fail.

This PR will work towards getting Travis working again for the `develop` branch, and then existing PRs can be updated to pull these fixes in and so hopefully pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+dist: xenial
+
+services:
+  - mysql
+
 language: php
 
 notifications:
@@ -8,6 +13,9 @@ notifications:
 php:
   - 5.6
   - 7.0
+  - 7.1
+  - 7.2
+  - 7.3
 
 env:
   - WP_VERSION=latest WP_MULTISITE=0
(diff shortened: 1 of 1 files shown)


GaryJones:
I'll squash and rebase and make this a proper PR. Nice to have them passing again.

GaryJones:
Look at [Travis](https://travis-ci.org/Automattic/Cron-Control/branches?utm_medium=notification&utm_source=github_status) to see that branches based off of `develop` broke sometime between 10 and 4 months ago (presumably when the Travis default changed from Trusty to Xenial).

Those four branches (two for me, one for @seanosh, one for @dchymko) should be rebased onto `develop` once this PR has been merged into `develop`, so that the tests will run.

Here are the full changes:

- Switch to Xenial
  We still need to test against PHP 5.6. PHP 5.6 is included with Xenial, but not with Bionic - and although the job tries to install it, it fails with:

   ```
  $ curl -sSf --retry 5 -o archive.tar.bz2 $archive_url && tar xjf archive.tar.bz2 --directory /
  curl: (22) The requested URL returned error: 404
  ```

  So, we're switching back to Xenial for now. Once we decide that PHP 5.6 and 7.0 is no longer needed, we can use Bionic instead.

- Set MySQL service
  This Travis config file was last amended when Trusty was the default Linux distribution for Travis. The default is now Xenial, and both Xenial and Bionic no longer include the MySQL service by default, and as such, the unit tests fail to initialise and all PRs fail.

- Download and run different PHPUnit versions
  See https://phpunit.de/supported-versions.html for which versions of PHP that PHPUnit supports.

- Test against PHP 7.1 through to PHP 7.3.
  We have to be compatible with WP Integration tests, so that means only be able to use PHPUnit 7, and that means only being able to test up to PHP 7.3.

GaryJones:
Noting that the master branch already has a Travis config that is significantly different, including a composer.json file which specifies PHP 7.0 as the minimum.

Closing this PR for now, until I can get this all consolidated.

GaryJones:
Fix for Travis config on `master`: https://github.com/Automattic/Cron-Control/pull/186

PR #188, opened 2020-04-01
Add command to list sites divided up by host

joshbetz:
Previously the runner would use `wp site list` to get a list of sites to
check. The problem is that if you have many runners (i.e. many cron hosts
per site) we waste a lot of time trying to run the same jobs on every host.

The idea here is to dedicate certain sites to certain hosts.

The runner will run `wp cron-control orchestrate sites list` to get a list
of sites. The act of running that command triggers a "heartbeat", which
stores a list of all hosts in the object cache along with the last
heartbeat time.

We use the list of hosts to divide sites into groups. Right now the idea is
that each site will have 2 hosts dedicated to running cron for it so that
if a host goes away, we have a bit of time to recover.

If a host is not seen for 3 heartbeats, it is removed from the list.

### DIFF
--- a/includes/wp-cli.php
+++ b/includes/wp-cli.php
@@ -79,4 +79,5 @@ require __DIR__ . '/wp-cli/class-lock.php';
 require __DIR__ . '/wp-cli/class-one-time-fixers.php';
 require __DIR__ . '/wp-cli/class-orchestrate.php';
 require __DIR__ . '/wp-cli/class-orchestrate-runner.php';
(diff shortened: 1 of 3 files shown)


joshbetz:
Did some testing in a sandbox.

* Everything works the same as before if `-smart-site-list` is not set, as expected.
* With 1 group everything works the same as before.
* With 2+ groups, the number of sites processed per container drops in a manner corresponding to the number of containers/groups.

The next step here is to deploy the new command, deploy the new runner (and the corresponding VIPd changes to enable that) and run it on a test site in production.

PR #189, opened 2020-04-09
Split the heartbeat from site retrieval

joshbetz:
Rather than inherently recording a heartbeat when the site list is 
retrieved, we'll do it when the runner records a heartbeat. The heartbeat
process is async whereas site list retrieval is actually blocked by running
events. For example, if events take longer than the event interval to run,
it will take longer than the event interval to get the list of sites. We
were seeing some issues in production where it was taking longer than 3x
the event interval to run events and hosts were being removed from
processing. Recording heartbeats completely async every heartbeat_interval
should resolve this.


joshbetz:
@darthhexx this should be ready to go. I can get the WP-CLI command deployed if this looks OK. There are a couple extra things in here that made testing locally easier, but I can remove them if you want. Namely, the mockcli to mock the necessary WP-CLI commands, more verbose errors when JSON parsing fails, and force exit immediately on the second SIGINT.

PR #194, opened 2020-07-20
Site List: Remove unnecessary switch_to_blog

joshbetz:
The `get_home_url()` function calls `switch_to_blog()` and `restore_current_blog()` internally.

First, `restore_current_blog()` is basically `switch_to_blog()` in reverse. In this context, we don't actually care about switching back. We only want this one option for every site, so switching back is a waste of time. After we get the option, we're just outputting it.

More importantly though, `switch_to_blog()` is expensive and not necessary. We're only using the home URL for the scheme, which is unnecessary for a "URL" that gets passed to WP-CLI. The domain + path is enough (i.e. `wp --url=example.com/subsite`). It also seems a little silly, in this context at least, to do a full `switch_to_blog()` to query this one database option. It seems like we just need to know the table name and from there could query the options table directly. The expensive part of `switch_to_blog()` is where we re-initialize the object cache, thus wiping out all auto-loaded options.

### DIFF
--- a/includes/wp-cli/class-orchestrate-sites.php
+++ b/includes/wp-cli/class-orchestrate-sites.php
@@ -60,11 +60,6 @@ class Orchestrate_Sites extends \WP_CLI_Command {
 			// Don't use home or siteurl since those don't always match up with the `wp_blogs` entry.
 			// That can result in a "site not found" when passed via the `--url` WP-CLI param.
 			// Instead, construct the URL from data in the `wp_blogs` table.
-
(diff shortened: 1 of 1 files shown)


joshbetz:
Also noting that this is only possible after https://github.com/Automattic/Cron-Control/pull/193, where we use the domain and path directly. We were previously using the `get_home_url()` output.

joshbetz:
r1042-deploy

darthhexx:
Thanks for taking care of this @joshbetz

PR #178, opened 2019-04-30
Support WP 5.1's new cron hooks

ethitter:
5.1 will introduce a new set of filters that we can use to intercept all cron-scheduling operations. 

Outstanding items:
- [ ] Introduce helper to replace `get_option( 'cron' )` calls with `Events_Store` methods.
- [x] Tests

Fixes #173 
Ports #174 from `ethitter/Cron-Control`

PR #195, opened 2020-09-01
a8c_cron_control_option: Don't bypass the local cache

joshbetz:
By bypassing the local PHP cache, we potentially make many more requests to memcached than we would otherwise. Given that this option is used so heavily, it can result in exponentially more load on memcached for this one option.

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -712,7 +712,7 @@ class Events_Store extends Singleton {
 	 * @return array|false
 	 */
 	private function get_cached_option() {
-		$cache_details = wp_cache_get( self::CACHE_KEY, null, true );
+		$cache_details = wp_cache_get( self::CACHE_KEY );
 
 		if ( ! is_array( $cache_details ) ) {
 			return false;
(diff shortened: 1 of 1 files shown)


joshbetz:
@ethitter it's been a long time, do you happen to remember why we were bypassing the local cache here originally?

ethitter:
Oh my... 🤔

I have vague recollections of race conditions and/or duplicate events. If memory serves, it came up in production when both web and CLI containers were interacting with cron. There might be some added details on a P2 somewhere, either in my offboarding notes or in one of the posts I wrote about this plugin. 

The cache bypass was introduced in #89 when I switched the plugin from a CPT to a custom table, which was largely done to fix duplicates and race conditions. In 8dd705e, the `wp_cache_get()` call was accompanied by this note, which somewhat confirms my recollection: `Use cached value for reads, except when we're unscheduling and state is important`.

That all said, `Events_Store::get_uncached_option()` is only called by `Events_Store::get_option()`, which has a local cache of its own thanks to #165. If that cache is working correctly, I would expect it and the local object cache to be invalid at the same time, requiring the call to memcached anyway. Regardless, two layers of local cache seems risky. 

It's probably time to come back to #178. Manipulating the cron array is where we've always had the most trouble, and we can dispense with that.

joshbetz:
Thanks, we were just talking about this and came to the same conclusion. Multiple processes saving to the database based on a potentially stale cache is risky. Agreed #178 is the way to go.

PR #196, opened 2020-09-10
Lower force publish missed scheduled posts to 2 minutes

joshbetz:
This job is a significant % of all jobs across VIP Go and likely causing more problems than it currently solves by delaying other jobs. Since this is just a backup method, 2 minutes should be plenty.

This is just a first step. We should look at how useful this job really is in practice. Does it still make sense to have this job scheduled on all sites? Should we have a lower default (5 minutes? 10 minutes?) that individual sites can override? Should the force schedule job be a separate plugin altogether?

### DIFF
--- a/includes/class-internal-events.php
+++ b/includes/class-internal-events.php
@@ -60,7 +60,7 @@ class Internal_Events extends Singleton {
 	private function prepare_internal_events() {
 		$internal_events = array(
 			array(
-				'schedule' => 'a8c_cron_control_five_minute',
+				'schedule' => 'a8c_cron_control_minute',
 				'action'   => 'a8c_cron_control_force_publish_missed_schedules',
 				'callback' => array( $this, 'force_publish_missed_schedules' ),
 			),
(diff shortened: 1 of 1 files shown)


nickdaugherty (approved):

mjangda:
> Does it still make sense to have this job scheduled on all sites?

If we can guarantee a high degree of reliability for the cron system (e.g. fix problems with backlogged jobs), the need for this backup job goes away.

joshbetz:
> If we can guarantee a high degree of reliability for on-time execution with the cron system

If jobs are delayed, wouldn't these also be delayed?

ethitter:
The internal jobs should only be delayed in a situation where a site has an insufficient number of cron runners or an excessive number of long-running client jobs, preventing the queue from being processed in a timely manner. Those same conditions apply to internal jobs delaying client jobs. The intent was for the internal jobs to run consistently but without interfering with client jobs, which seems no longer to be the case.

PR #197, opened 2020-11-04
Don't schedule internal events in REST context

joshbetz:
Every call to wp_next_scheduled is a request to the object cache for the full cron option. We have 4 internal events, so that's 4 additional object cache requests on every REST API request. Limiting cron scheduled, as much as possible, to wp-admin will reduce load on our object cache.


mjangda (approved):
👍

WPprodigy:
Do we know why this was originally added? Best I can think of is for situations where a site is fully decoupled and the WP dashboard is never used - so `admin_init` would not be reliable to ensure schedules.

The trade-off on losing that and gaining this performance definitely seems acceptable though.

joshbetz:
@WPprodigy Yep, I think that's right and agree.

PR #208, opened 2021-08-16
Runner: Support PHP-FPM for wp-cli invocation

joshcoughlan:
1. You can now pass `-fpm-url=unix:///var/run/fastcgi.sock` and the runner will invoke a shim over FastCGI that runs WP-CLI in the fcgi context. This makes opcache work, so we do not recompile WordPress from PHP sources to bytecode every single time.
2. For comparison, a simple `wp option get home` takes 270ms over FPM, and 550ms using naïve wp-cli.
3. NOTE: `remote.go` still invokes wp-cli directly, so we do not need to worry about compatibility of _every conceivable_ wp-cli command with the shim.

If you do not pass the `-fpm-url` argument, the current behavior is unchanged (the runner will shell out to wp-cli for everything)


runner/runner.go
```diff
@@ -15,6 +15,8 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+	
+	"github.com/yookoala/gofast"
```
bnu0, line 19: can we use `github.com/kanocz/fcgi_client` here, so it shares dependencies with https://github.com/Automattic/php-fpm-exporter

runner/runner.go
```diff
@@ -110,7 +124,7 @@ func main() {
 	gEventRetrieversRunning = make([]bool, numGetWorkers)
 	gEventWorkersRunning = make([]bool, numRunWorkers)
 
-	if gMetricsListenAddr != "" {
+	if !fpm && gMetricsListenAddr != "" {
```
bnu0, line 127: why do we disable metrics when using fpm?

runner/runner.go
joshcoughlan: That was going to be my initial approach, but it doesn't look like that project is still active. There was another reason I decided to go with this other fgci client implementation, but I'll have to take another look at fcgi_client to remember what that reason was.

runner/runner.go
joshcoughlan: My impression was that if fpm is being used we wouldn't be getting any useful metrics anyways. If we want to allow it to listen and grab whatever metrics are getting updated I'm ok with that.

includes/fpm-cron-runner.php
```diff
@@ -0,0 +1,53 @@
...
+	$GLOBALS['argv'] = $cli_args;
+	// phpcs:ignore
+	$_SERVER['argv'] = $cli_args;
+
+	define( 'WP_CLI_ROOT', 'phar:///usr/local/bin/wp.phar/vendor/wp-cli/wp-cli' );
+	define( 'STDIN', fopen( 'php://input', 'r' ) );
+	define( 'STDOUT', fopen( 'php://output', 'w' ) );
+	define( 'STDERR', fopen( 'php://output', 'w' ) );
+
+	require_once WP_CLI_ROOT . '/php/wp-cli.php';
```
bnu0, line 52: You can require directly from a phar, using something like:

```php
require_once('phar://path/to/wpcli.phar/directory/inside/file.php')
```

includes/fpm-cron-runner.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+//
+// This file is being templated and managed by Kubernetes wpvip-operator.
+//
+
+//
+// An init container will be needed that runs something along the lines of
```
bnu0, line 8: We can put the wpcli phar inside the chroot in the image build even. Or extract it, but see my other comment regarding that.

runner/runner.go
```diff
@@ -83,6 +88,8 @@ func init() {
 	flag.StringVar(&logDest, "log", "os.Stdout", "Log path, omit to log to Stdout")
 	flag.StringVar(&logFormat, "log-format", "JSON", "Log format, 'Text' or 'JSON'")
 	flag.BoolVar(&debug, "debug", false, "Include additional log data for debugging")
+	flag.BoolVar(&fpm, "fpm", false, "Use FPM to run jobs")
+	flag.StringVar(&fpmUrl, "fpm-url", "unix:///var/run/fastcgi.sock", "Url for the php-fpm server or socket")
```
bnu0, line 92: I would just default this to the empty string `""` and providing a value here also enables fpm, so we do not need a separate book variable. This is fairly standard golang, using the "zero" value of a type to represent absence.

runner/runner.go
```diff
@@ -464,6 +473,78 @@ func runEvents(workerID int, events <-chan event) {
...
+		kv := strings.Split(s, "=")
+		k := kv[0]
+		v  := ""
+		if len(kv) == 1 && strings.HasPrefix(k, "--") {
+			queryData.Add(kv[0], "true")
+		} else if len(kv) == 1 {
+			queryData.Add("subcommands", kv[0])
+		} else if len(kv) == 2 {
+			queryData.Add(kv[0], kv[1])
+		} else if len(kv) > 2 {
```
bnu0, line 541: You can use `strings.SplitN` to avoid this case

runner/runner.go
```diff
@@ -464,6 +473,78 @@ func runEvents(workerID int, events <-chan event) {
...
+		path = fpmUrl.Fragment
+	}
+	
+	if fpmUrl.Scheme == "unix" {
+		host = fpmUrl.Path
+	}
+	
+	fcgi, err := fcgiclient.Dial(fpmUrl.Scheme, host)
+	if err != nil {
+		log.Printf("fastcgi Dial failed: %s\n", err)
```
bnu0, line 499: I think we need to `return "", err` here and a bunch of similar ones below...

runner/runner.go
```diff
@@ -464,6 +473,78 @@ func runEvents(workerID int, events <-chan event) {
...
+	env["WP_CLI_ROOT"] = "/usr/local/wp-cli"
+	// This script needs to be injected into $WP_CLI_ROOT by wpvip-operator
+	env["SCRIPT_FILENAME"] = "/var/www/wp-content/mu-plugins/cron-control/includes/fpm-cron-runner.php"
+	env["GATEWAY_INTERFACE"] = "FastCGI/1.0"
+	env["REQUEST_METHOD"] = "POST"
+	env["CONTENT_TYPE"] = "application/x-www-form-urlencoded"
+	return env
+}
+
+func buildFpmQuery(subcommand []string) string {
```
bnu0, line 529: It feels like it would just be easier to Json encode this array and decode it in php, than having to mess with posting a proper form...

includes/fpm-cron-runner.php
joshcoughlan: You can, but WP_CLI_ROOT is used elsewhere in the cli and so pointing that to the proper path within the phar I thought was best.

runner/runner.go
joshcoughlan: Ok. will do

runner/runner.go
```diff
@@ -464,6 +473,78 @@ func runEvents(workerID int, events <-chan event) {
...
+	
+	if path == "" || fpmUrl.Scheme == "unix" {
+		path = fpmUrl.Fragment
+	}
+	
+	if fpmUrl.Scheme == "unix" {
+		host = fpmUrl.Path
+	}
+	
+	fcgi, err := fcgiclient.Dial(fpmUrl.Scheme, host)
```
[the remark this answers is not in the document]
joshcoughlan: erm... yeah. Good catch.

runner/runner.go
joshcoughlan: I don't see how that would do what this is doing. This is distinguishing between positional args, flags, and kwargs.

runner/runner.go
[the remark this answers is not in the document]
joshcoughlan: done

runner/runner.go
joshcoughlan: Done... I think

runner/runner.go
joshcoughlan: I don't really have a good answer for you on this one. I thought about it, but I'm just not sure how to preserve the positional args in json. I guess maybe nesting the subcommands (positionals)? It doesn't seem like it would be a whole lot better to me. I think the problem is more on the receiving end and that it is a cli being hacked on.

includes/fpm-cron-runner.php
joshcoughlan: I think you may have posted this before I changed it to use the phar directly and I didn't see it?

bnu0 (approved):
LGTM 😎

joshcoughlan:
LGTM

PR #210, opened 2021-08-31
switch back to gofast

bnu0:
Unfortunately, `tomasen/fcgi_client` does not support splitting out the fcgi stderr stream from the stdout. On a well-behaved site that does not spew notices, thats fine. On most real sites, it's not fine.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -122,7 +122,7 @@ func init() {
 }
 
 func main() {
-	logger.Printf("Starting with %d event-retreival worker(s) and %d event worker(s)", numGetWorkers, numRunWorkers)
+	logger.Printf("Starting with %d event-retrieval worker(s) and %d event worker(s)", numGetWorkers, numRunWorkers)
 	logger.Printf("Retrieving events every %d seconds", getEventsInterval)
 	go setupSignalHandler()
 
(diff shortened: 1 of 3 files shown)


joshcoughlan:
Any reason not to use connection pooling? Is the benefit just not worth the added complexity?

runner/runner.go
```diff
@@ -491,55 +492,76 @@ func runWpFpmCmdWithMetrics(subcommand []string) (string, error) {
...
+}
 
-	fcgi, err := fcgiclient.Dial(fpmUrl.Scheme, host)
+func (f *fakeHttpResponseWriter) WriteHeader(statusCode int) {
+	f.LastStatus = statusCode
+}
+
+func runWpFpmCmdSafe(subcommand []string) (string, error) {
+
+	fpmClient, err := fpm()
```
bnu0, line 518: Resource leak, client is Closeable!

bnu0:
> Any reason not to use connection pooling? Is the benefit just not worth the added complexity?

If this was using TCP to a remote process, i would, but it's using a unix domain socket and the overhead of connecting to it repeatedly is probably less than any theoretical connection pool implementation.

joshcoughlan (approved):
LGTM

PR #213, opened 2021-11-10
Add new --queue-size & --queue-window args to the orchestrator's "list-due-batch" CLI

WPprodigy:
## Description

To go alongside the new cron runner, we need to be able to customize (or rather disable for now) the functionality that fetches jobs that are not currently "due now". 

Don't want to hardcode a change to `JOB_QUEUE_SIZE` right now, as the new runner is going to be intended for use on k8s and not VIPd - thus PHP changes must remain fully BC.

## Testing

(use `wp cron-control events list` to keep an eye on expected results)

```
// Lists events due up to +60s into the future
wp cron-control orchestrate runner-only list-due-batch

// List only events that are "due now"
wp cron-control orchestrate runner-only list-due-batch --queue-window=0

// List events that are due in the next 5 minutes.
wp cron-control orchestrate runner-only list-due-batch --queue-window=300
```

____

And to allow future tailoring in the runner, I also added a flag to change the queue size.

```
// Fetches only up to (JOB_QUEUE_SIZE + internal) events
wp cron-control orchestrate runner-only list-due-batch --queue-window=900000

// Fetches up to (2 + internal) events
wp cron-control orchestrate runner-only list-due-batch --queue-size=2

// Likely fetches all events
wp cron-control orchestrate runner-only list-due-batch --queue-size=1000 --queue-window=900000 --format=count
```

### DIFF
--- a/includes/class-events.php
+++ b/includes/class-events.php
@@ -91,7 +91,7 @@ class Events extends Singleton {
 	/**
 	 * List events pending for the current period
 	 *
-	 * @param array $job_queue_size   Maximum number of events to return.
+	 * @param array $job_queue_size   Maximum number of events to return (excludes internal events).
 	 * @param array $job_queue_window How many seconds into the future events should be fetched.
 	 * @return array
 	 */
@@ -188,7 +188,7 @@ class Events extends Singleton {
 	}
 
 	/**
-	 * Trim events queue down to the limit set by JOB_QUEUE_SIZE
+	 * Trim events queue down to a specific limit.
 	 *
 	 * @param array $events         List of events to be run in the current period.
 	 * @param array $max_queue_size Maximum number of events to return.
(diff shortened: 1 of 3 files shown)


jesse-greathouse (commented):
I like the functionality around adding a maximum job queue size and being able to override the global settings with the command line args/extra parameters. Im not sure what spawned this PR but Im betting something is getting hosed by job queue events?

I wish there was some testing around whether feeding it these extra parameters will verify that the global values will not be used in that case.

WPprodigy:
Oi, the testing in this plugin could really use some love, setup was not super easy since it doesn't use local dependencies.

Definitely worth taking a deeper look at when we look into adding support for WP's new cron hooks.

ethitter:
> ...starting on adding support for WP's new cron hooks

FYI, that's mostly done already: https://github.com/Automattic/Cron-Control/pull/178

jesse-greathouse (approved):
LGTM

WPprodigy:
@ethitter Yep! Have been eyeing that for a while :). Time permitting, definitely hoping to get around to finishing it out.

PR #214, opened 2021-11-17
Add GitHub workflows

sjinks:
This is a pretty big PR that fixes the CI workflow and switches to GitHub Actions from Travis. It should probably have been split into several ones, but one change requires another, and it was pretty hard without the working CI.

The changes are:
* add missing dependencies to `package.json`;
* this forced to update outdated dependencies; as a side effect, this fixes the CVE-2020-7729 in `grunt`;
* add `package-lock.json` to the version control so that we can run reproducible tests;
* add test workflows for Grunt (run all Grunt tasks; for that, I had to fix `Gruntfile.js` a bit) ~~and Go (build the runner; again, this required me to modify the Makefile a bit)~~;
* as a result, regenerate `README.md` and `languages/cron-control.pot`;
* fix issues found by PHPCS;
* add `composer.lock` to the version control for the sake of reproducible builds;
* add several developer dependencies to improve the DX (so that we can run tests, phpcs, etc locally).


.github/workflows/ci-go.yaml
```diff
@@ -0,0 +1,22 @@
...
+    if: "!contains(github.event.head_commit.message, '[ci skip]')"
+    name: Build Runner
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2.3.4
+
+      - name: Build Runner
+        run: make
+        working-directory: runner
```
WPprodigy, line 22: Oh, we can actually scrap any runner-related stuff in this. It's being moved to https://github.com/Automattic/cron-control-runner and the build process will be different than seen here.

README.md
```diff
@@ -18,7 +18,7 @@ Using REST API endpoints (requires WordPress 4.4+), or a Golang daemon, an event
 
 ## Installation ##
 
-1. Define `WP_CRON_CONTROL_SECRET` in `wp-config.php`, set to `false` to disable the REST API interface.
+1. Define `WP_CRON_CONTROL_SECRET` in `wp-config.php`
```
WPprodigy, line 21: I think we should leave this, helps guide the setup a bit better.

sjinks: It was regenerated by Grunt from readme.txt, I believe - probably need to update it there

.github/workflows/ci-go.yaml
sjinks: @WPprodigy just to confirm, can I remove the entire `runner` directory?

WPprodigy: Not yet, I have a deprecation plan. But just don't worry about any improvements or building.

README.md
WPprodigy: Ahh, right. I'll handle that.

.github/workflows/ci-go.yaml
sjinks: OK

README.md
sjinks: I already did that

.github/workflows/ci-go.yaml
sjinks: Done

WPprodigy (approved):
This is so awesome, thanks a ton for doing this <3

WPprodigy:
Removed travis check from the branch protection rules.

PR #219, opened 2021-11-24
Implement WP cron filters

WPprodigy:
Refreshed work that continues w/ what started in #178. Thanks immensely to @ethitter & @joshbetz for the prior art on this.

## Description

This implements the filters that were added in WP 5.1 that allow us to intercept every core cron event scheduling operation. As a result, we no longer have to rely on one massive `cron` option that contains everything (race-conditions begone!). So the first section of filters you see here allow us to properly use the custom events storage directly when WP cron-related functions are called.

The second section of filters are re-implementations of existing logic, just in case a plugin for example is reading from the `cron` option, and maybe even deleting/adding events directly. So for full BC we still support doing that.

## Notable changes

(Some of this will be more visible in the other PRs)

1. Clean interaction layer for events. `Event::get()`, `$event->complete()`, etc. The logic is enclosed in that object.
2. One other deviation from the prior `#178` PR is that caching for events is handled individually rather than sharing the same `cron_control_jobs` cache key. This will help prevent race conditions.
3. Returns WP_Error's on failure. Core will automatically translate this to `false` pending on if the requester asked for a WP_Error or not.
4. In `pre_schedule_event`, I deviated from core slightly. It will reject scheduling a duplicate recurring event (one already exists that is active and has the same schedule), as this has caused many problems in the past. Interestingly, [core docblocks](https://developer.wordpress.org/reference/functions/wp_schedule_event/) say that there is some duplication prevention for recurring events, but there is in fact none.
5. The `pre_get_cron_option` avoids the cache bucketing issues we previously had, as it uses the new query layer that automatically caches things based on SQL args. W/o going into too much details on that (will cover in a separate PR), suffice to note that this will help immensely if anybody is still out there asking for the full cron array.
6. I leaned into the "complete" status, and opted for using it rather than directly deleting. Reasoning here is that I'd like to do more w/ statuses in the future and follow a path similar to Action Scheduler. Where we cleanup (truly delete) events that are older than X days, but leave them at say a completed/cancelled/failed status for a few days along with potentially have plugin-level logging with the events so failures can be better seen.

## Testing

You'll actually want to checkout the `next/complete-changeset` branch to test the full picture together currently. Once checked out, can run unit tests and various WP flows where events are added/ran. I recommend reviewing the new file locally w/ the complete changeset branch so you can refer to other methods.

**Expect tests to fail in the individual PRs.** The final PR will be the passing one :)

## PR Process Notes
This is one of a few parts of changes that I have currently to ship together. To help keep things "small" and reviewable, I've created a branch called `next/pull-request-base` that these PRs will be opened against. As they are reviewed and approved, I'll merge into this branch. Everything going into `next/pull-request-base` will be approved via PR, I've set up a temporary branch protection on it.

Once all the pieces are in, `next/pull-request-base` will be rebased against master and can become the new "cron control next" branch for pre-prod testing, eventually to be merged into master for the production release.

### DIFF
--- a/includes/wp-adapter.php
+++ b/includes/wp-adapter.php
@@ -1,0 +1,314 @@
+<?php
+
+namespace Automattic\WP\Cron_Control;
+
+use Automattic\WP\Cron_Control\Events_Store;
+use WP_Error;
+
+// Integrate w/ WP's cron api once the custom table is installed.
+if ( Events_Store::is_installed() ) {
+	// Core filters added in WP 5.1, allowing us to fully use the custom event's store.
+	add_filter( 'pre_schedule_event', __NAMESPACE__ . '\\pre_schedule_event', 10, 2 );
+	add_filter( 'pre_reschedule_event', __NAMESPACE__ . '\\pre_reschedule_event', 10, 2 );
+	add_filter( 'pre_unschedule_event', __NAMESPACE__ . '\\pre_unschedule_event', 10, 4 );
+	add_filter( 'pre_clear_scheduled_hook', __NAMESPACE__ . '\\pre_clear_scheduled_hook', 10, 3 );
+	add_filter( 'pre_unschedule_hook', __NAMESPACE__ . '\\pre_unschedule_hook', 10, 2 );
+	add_filter( 'pre_get_scheduled_event', __NAMESPACE__ . '\\pre_get_scheduled_event', 10, 4 );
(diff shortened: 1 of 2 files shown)


pschoffer (approved):
This looks great. There are some things that would change here after the previous PR feedback is resolved (if any changes), but over all looks great.

WPprodigy:
Thanks. I actually don't think any of the other changes overlap here which is pretty cool if so, but if they do I'll address in an additional PR or perhaps with the said changes.

So going to go ahead and merge this so we can work towards smaller diff sizes xD

PR #221, opened 2021-11-24
New event's store methods

WPprodigy:
## Description

Here are the aforementioned four main new methods on the data store:

```
_create_event( array $row_data ): int
_update_event( int $event_id, array $row_data ): bool
_get_event_raw( int $id ): ?object
_query_events_raw( array $args = [] ): array
```

These do pretty much all the heavy lifting needed for querying. In fact, the "working branch" already has additional cleanup in this class that I'll be PRing afterwards that replaces other store methods to internally use these new ones. That PR will look a lot more pleasant, sitting at `+110, -521` :), and has full backwards compatibility w/ all currently public methods and existing unit tests.

Some notes about the PR:

- Cache invalidation is handled by the `last_changed` cache group key. Whenever an update occurs, this is incremented to the current timestamp, signaling the query class that it needs to get fresh data. This works essentially the same way the old way of caching the entire cron option array in multiple buckets, except now they are individual entries w/o concerns for race conditions occurring while the cache is built up.
- A future enhancement I'm eyeing is caching individual events on a `action/instance/timestamp` key w/ their own group. Since the main action that happens on the FE is checking if an event is already registered (w/ these 3 args), this will help keep invalidations to a minimum for those queries as we can invalidate them specifically only when that event is changed. I've already planned out how this will work and we can do it w/ full BC later if we find it's needed.
- I'm not a huge fan of the "fake private" static methods, but it will take a few rotations of deprecations before we can make Events_Store a non-singleton and better enclose it within the plugin. So for now this seemed best.
_____

## Testing

You'll actually want to checkout the `next/complete-changeset` branch to test the full picture together currently. I recommend doing that locally while reviewing this new file rather than just via GH.

**Expect tests to fail in the individual PRs.** The final PR will be the passing one :)

## PR Process Notes
This is one of a few parts of changes that I have currently to ship together. To help keep things "small" and reviewable, I've created a branch called `next/pull-request-base` that these PRs will be opened against. As they are reviewed and approved, I'll merge into this branch. Everything going into `next/pull-request-base` will be approved via PR, I've set up a temporary branch protection on it.

Once all the pieces are in, `next/pull-request-base` will be rebased against master and can become the new "cron control next" branch for pre-prod testing, eventually to be merged into master for the production release.

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -961,7 +961,7 @@ class Events_Store extends Singleton {
 
 	/**
 	 * Get raw events data based on various available query args.
-	 * For internal use only, please use Event::get( $args ) or Events::query( $args ).
+	 * For internal use only, please use Event::find( $args ) or Events::query( $args ).
 	 *
 	 * @param array $args Argument list for the query.
 	 * @return array Array of raw event objects.
@@ -1012,18 +1012,7 @@ class Events_Store extends Singleton {
 			],
 		];
 
-		$parsed_args = wp_parse_args( $args, [
-			'action'        => $valid_args['action']['default'],
-			'action_hashed' => $valid_args['action_hashed']['default'],
-			'args'          => $valid_args['args']['default'],
-			'instance'      => $valid_args['instance']['default'],
-			'timestamp'     => $valid_args['timestamp']['default'],
-			'schedule'      => $valid_args['schedule']['default'],
-			'status'        => $valid_args['status']['default'],
-			'limit'         => $valid_args['limit']['default'],
-			'page'          => $valid_args['page']['default'],
-			'order'          => $valid_args['order']['default'],
-		] );
+		$parsed_args = wp_parse_args( $args, array_map( fn( $arg ) => $arg['default'], $valid_args ) );
 
 		foreach ( $valid_args as $arg_name => $arg_checks ) {
 			if ( $parsed_args[ $arg_name ] !== $arg_checks['default'] ) {
@@ -1067,7 +1056,7 @@ class Events_Store extends Singleton {
 		}
 
 		if ( ! is_null( $parsed_args['action_hashed'] ) ) {
-			// TODO: Deprecate this query arg later once all is converted to the new API.
+			// TODO: Deprecate this query arg later once all is converted.
 			$sql .= ' AND action_hashed = %s';
 			$placeholders[] = $parsed_args['action_hashed'];
 		}
@@ -1078,7 +1067,7 @@ class Events_Store extends Singleton {
 			$sql .= ' AND instance = %s';
 			$placeholders[] = $instance;
 		} elseif ( ! is_null( $parsed_args['instance'] ) ) {
-			// TODO: Deprecate this query arg later once all is converted to the new API.
+			// TODO: Deprecate this query arg later once all is converted.
 			$sql .= ' AND instance = %s';
 			$placeholders[] = $parsed_args['instance'];
 		}
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -27,11 +27,8 @@ class Events_Store extends Singleton {
 	const STATUS_PENDING   = 'pending';
 	const STATUS_RUNNING   = 'running';
 	const STATUS_COMPLETED = 'complete';
-	const ALLOWED_STATUSES = array(
-		self::STATUS_PENDING,
-		self::STATUS_RUNNING,
-		self::STATUS_COMPLETED,
-	);
+	const ACTIVE_STATUSES  = [ self::STATUS_PENDING, self::STATUS_RUNNING ];
+	const ALLOWED_STATUSES = [ self::STATUS_PENDING, self::STATUS_RUNNING, self::STATUS_COMPLETED ];
 
 	const CACHE_KEY = 'a8c_cron_ctrl_option';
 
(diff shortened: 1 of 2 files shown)


includes/class-events-store.php
```diff
@@ -882,6 +879,315 @@ public function count_events_by_status( $status ) {
...
+
+	/**
+	 * Update an event.
+	 * For internal use only, please use Event::save() as this does not validate.
+	 *
+	 * @param int   $event_id The ID of the event being updated.
+	 * @param array $row_data The row data used to update the event.
+	 * @return bool True if update was successful, false otherwise.
+	 */
+	public static function _update_event( int $event_id, array $row_data ): bool {
```
pschoffer, line 921: Same as above... 

or if we want to abstract away DB operations from event class. Let's make a new class like `EventRepository` and inject it to Event class. 

That might actually work even better and decouple dependencies.

You could have `Event( $eventRepository )` then:
```
class EventFactory {
  EventRepository $eventRepository
  
  __constructor() {
     $eventRepository = new EventRepository();
  }

   public createEvent() {
      return new Event($eventRepository);
   }
   
  public getFromEventData($eventData) {
      $event = $this->createEvent();
      $event->hidrate($eventData);
      return $event;
   }
   
   public getFromId($id) {
      $eventData = $eventRepository->get($id);
      return $this->getFromEventData($eventData);
   }
}
```

The benefit is that it is all decoupled. You can test the repository without Event object. And event object with the mocked repository.

We can of course name `EventFactory` `EventStore` - that is this class, if that makes more sense.

includes/class-events-store.php
```diff
@@ -882,6 +879,315 @@ public function count_events_by_status( $status ) {
...
+				'default'    => 1,
+				'validation' => fn( $page ) => is_int( $page ) && $page >= 1,
+			],
+			'order' => [
+				'default'    => 'ASC',
+				'validation' => fn( $order ) => is_string( $order ) && in_array( strtoupper( $order ), [ 'ASC', 'DESC'], true ),
+			],
+		];
+
+		$parsed_args = wp_parse_args( $args, [
```
pschoffer, line 1015: This repeated list could be replaced with:
```
array_map(function( $argument ) { return $argument['default'] }, $valid_args );
```

Or even with:
```
array_map(function( $argument ) { return $argument['default']  ?? null }, $valid_args );
```
and then drop the lines setting `null` as defaults (most args) from above.

includes/class-events-store.php
```diff
@@ -882,6 +879,315 @@ public function count_events_by_status( $status ) {
...
+			if ( ! is_null( $parsed_args['page'] ) ) {
+				$offset = $parsed_args['limit'] * ( $parsed_args['page'] - 1 );
+				if ( $offset > 0 ) {
+					$sql .= ' OFFSET %d';
+					$placeholders[] = $offset;
+				}
+			}
+		}
+
+		$last_changed = wp_cache_get_last_changed( 'cron-control-events' );
```
pschoffer, line 1121: You were right! This is kinda smart! :)

pschoffer (changes requested):
Left a big refactoring suggestion and some minor nitpicks. The logic itself seems great (I like no surprises in code)

includes/class-events-store.php
WPprodigy: Somehow I never picked up on the core patterns that use this. Makes cache groups a lot more useful. https://github.com/WordPress/WordPress/blob/121c824ccde0a020f2e92b6d287d3a6fc35a913f/wp-includes/post.php#L5630-L5634

includes/class-events-store.php
WPprodigy: Ah smart, I looked past the duplication here :). I'll go w/ the first I think and leave `null` defaults, as I kinda like how it helps document usage automatically in `$valid_args`.

includes/class-events-store.php
```diff
@@ -882,6 +879,315 @@ public function count_events_by_status( $status ) {
...
+	*/
+
+	/**
+	 * Create an event.
+	 * For internal use only, please use Event:save() as this method does not validate.
+	 *
+	 * @param array $row_data The row data used to create the event.
+	 * @return int The newly created event ID, 0 if creation failed.
+	 */
+	public static function _create_event( array $row_data ): int {
```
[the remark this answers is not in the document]
WPprodigy: I like keeping the SQL together, as it also keeps the caching/invalidating in one spot.

includes/class-events-store.php
```diff
@@ -882,6 +879,315 @@ public function count_events_by_status( $status ) {
 
 		return (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(ID) FROM {$this->get_table_name()} WHERE status = %s", $status ) ); // Cannot prepare table name. @codingStandardsIgnoreLine
 	}
+
+	/*
+	|--------------------------------------------------------------------------
+	| New event's store methods. The above may be deprecated in the future.
+	| But notably, the below is also internal-usage only. See comments about alternatives.
```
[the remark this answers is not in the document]
WPprodigy: "internal" meaning just within the "cron control" plugin internals specifically, not even the rest of mu-plugins. Any caller should be going through Event/Events.

Would definitely be ideal to be truly private, but kind of hard to accomplish within the existing Singleton structure. In the future once I'm able to deprecate out the public methods in the singleton, I envisioned something more strict. At which point I'd be pretty comfortable dropping these "fake private" methods since it was pretty clear one was not supposed to use these.

includes/class-events-store.php
WPprodigy: A few things come to mind:

1) It is a bit heavy on the added complexity side, though there are benefits of course.
2) Real unit testing is made easier w/ the factory, but WP Unit tests are more-so integration tests which is fine but as a result we don't get much benefit since we might as well test the full route.
3) It would be initializing a lot more objects when dealing w/ thousands of `Event`s, though this is probably a non-issue.
4) Not totally sure where things like `STATUS_PENDING` const would belong in the new structure long-term.

Was thinking maybe could cut out the factory middleman initially, and go w/:

```
class Event {
  private EventRepository $eventRepository

  __constructor() {
     $this->eventRepository = new EventRepository();
  }

   public function save() {
      ... logic ...
      return $this->eventRepository->update( $args );
   }
}
```

It keeps the logic enclosed so we'd be free to change it. And if in the future we ever wanted to support custom data stores, then adding the factory layer could make sense - providing a filter essentially to swap out the store for anther that provides the same interface.
____

But even w/ the above, I think we still end up with the same problem don't we? `EventRepository` is effectively what these three "fake private" methods are in Events_Store right now, just w/ extra steps. They have to live somewhere together, and they'd ultimately be accessible?

pschoffer: Yeah I like that... the injecting in the constructor is what I was after. `Factory` is an afterthought. 

> EventRepository is effectively what these three "fake private" methods are in Events_Store right now

I think in this scenario EventRepository is EventStore so no extra step. Also, I think it is perfectly fine to have them public then. This fake_privatenes  + the fact that they are only used at one place (so dont need to be public if we would like to enforce them being private) is what confused me in the first place. :) 

I mean it is fine to expose the registry functionality so that people dont have to write their own SQL if they would like to skip the Event abstraction.

Thinking about it some more... maybe the issue here is that we are mixing the abstraction layer that the user should interact with (the EventStore/Event) together with the lower level Data Access layer that we don't want the user to interact with. Making sort of cyclic dependency. `EventStore` -> `Event` -> `EventStore (for SQL stuff)`. If we move out the SQL stuff to a separate class it might be cleaner `EventStore` -> `Event` -> `EventRepository` and `EventStore` -> `EventRepositoy`.

What do you think?

WPprodigy: Summarizing our chat for historical record:

- There won't be circular dependencies in the end result, it will go just from Event or Events -> EventsStore. In the interim there may be some circles while we deprecate methods in the store, but the end result will be linear. The store is the last stop, handling the SQL and caching layers.

- As such, the end result is that the consumer/user will never be calling the EventStore methods at all once everything is in place. Events and Event will hold wrappers for any SQL-related functionality so we can ensure it goes through the right paths.

WPprodigy:
Applied the nifty array map for query arg parsing/defaults. Leaving the rest as-is after further discussion, but notably nothing is really being set in stone here as it will be possible to adjust in the future without much fear of backwards-compatibility breaks.

pschoffer (approved):
:on:

PR #223, opened 2021-11-30
Don't alter JSON parameters on WP-CLI commands

dchymko:
## Description 
This change fixes a bug that would strip double quotes from the property name and values inside a JSON payload. While not a "failure" bug, this would cause unexpected consequences for some clients when the values weren't stored with the quotes.
`vip @site.env --yes -- post meta update 1 custom_option '{\"name\":\"Text\",\"version\":0}'`

It also fixes a bug where if there was a space in a JSON value, the command would fail. THis command should now work:
`vip @site.env --yes -- post meta update 1 custom_option '{\"name\":\"Some text with spaces\",\"version\":0}'`

### DIFF
--- a/runner/remote.go
+++ b/runner/remote.go
@@ -332,7 +332,7 @@ func validateAndProcessCommand(calledCmd string) (string, error) {
 }
 
 func getCleanWpCliArgumentArray(wpCliCmdString string) ([]string, error) {
-	rawArgs := strings.Fields(wpCliCmdString)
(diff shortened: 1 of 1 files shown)


bnu0:
This is no longer the runner used on k8s, which can now be found at https://github.com/Automattic/cron-control-runner

dchymko:
> This is no longer the runner used on k8s, which can now be found at https://github.com/Automattic/cron-control-runner

Ah, I didn't realize that, thanks!

PR #224, opened 2021-12-06
New Event class methods

WPprodigy:
An additional part of the "cron control next" project, noting that this PR is going into `next/pull-request-base`. Just three new methods here:

- `Events::query( $query_args )` as a helper for querying multiple events at once.
- And then `format_events_for_wp()` and `flatten_wp_events_array()` are utilities for working w/ the annoyingly nested array format that WP uses in the `cron` option by default.

Definitely expecting some more methods here later, as anything that deals w/ multiple `Event`s will be living in this object. So in the future, things like bulk deletions will take place here as well. But I figure it'll be best to add those in during the particular cleanups where they are needed/introduced.

As for tests in this PR, I renamed the file since it didn't follow the convention the rest of the tests use (though I hope to maybe change this later). Mentioning this so that during review you can just click on the second commit to see the actual changes: https://github.com/Automattic/Cron-Control/commit/6643a8f7ddc2dfd24da02614cec6fa73dbddd053


includes/class-events.php
```diff
@@ -559,6 +559,99 @@ public function update_run_status( $new_status ) {
...
 	}
+
+	/**
+	 * Query for multiple events.
+	 *
+	 * @param array $query_args Event query args.
+	 * @return array An array of Event objects.
+	 */
+	public static function query( array $query_args = [] ): array {
+		$events  = Events_Store::_query_events_raw( $query_args );
```
pschoffer, line 570: Could we rename to `$event_db_rows` to make it clear that is NOT a collection ov Event objects?

includes/class-events.php
```diff
@@ -559,6 +559,99 @@ public function update_run_status( $new_status ) {
...
+		$objects = array_map( fn( $event ) => Event::get_from_db_row( $event ), $events );
```
pschoffer, line 571: Now this should be $events I think.

includes/class-events.php
```diff
@@ -559,6 +559,99 @@ public function update_run_status( $new_status ) {
...
+				'args'     => $event->get_args(),
+			];
+
+			$interval = $event->get_interval();
+			if ( ! empty( $interval ) ) {
+				$formatted_event['interval'] = $interval;
+			}
+
+			$instance = $event->get_instance();
+			$crons[ $timestamp ][ $action ][ $instance ] = $formatted_event;
```
pschoffer, line 609: oh this structure is lovely :)

pschoffer (changes requested):
Looks good. I left few minor naming notes

includes/class-events.php
```diff
@@ -559,6 +559,99 @@ public function update_run_status( $new_status ) {
...
+
+	/**
+	 * Query for multiple events.
+	 *
+	 * @param array $query_args Event query args.
+	 * @return array An array of Event objects.
+	 */
+	public static function query( array $query_args = [] ): array {
+		$event_db_rows = Events_Store::_query_events_raw( $query_args );
+		$events = array_map( fn( $event ) => Event::get_from_db_row( $event ), $event_db_rows );
```
pschoffer, line 571:
```suggestion
		$events = array_map( fn( $db_row ) => Event::get_from_db_row( $db_row ), $event_db_rows );
```

WPprodigy: Done in https://github.com/Automattic/Cron-Control/commit/b9b876fb4d14ad067687c52189a677dd976d8a06

PR #225, opened 2021-12-06
Misc cleanup for "cron control next" changeset

WPprodigy:
Helps tie the previous `next/pull-request-base` PRs together.

- Swap usage to the newer `Event::find()` method.
- Require new files.
- Add new test utils & fix testing bootstrap.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -37,13 +37,18 @@ function _manually_load_plugin() {
 	require dirname( dirname( __FILE__ ) ) . '/cron-control.php';
 
 	// Plugin loads after `wp_install()` is called, so we compensate.
-	// See the `class_init()` method in Events_Store for the logic behind this.
 	\Automattic\WP\Cron_Control\Events_Store::instance()->prepare_table();
-	remove_filter( 'schedule_event', '__return_false' );
-	add_filter( 'pre_option_cron', array( \Automattic\WP\Cron_Control\Events_Store::instance(), 'get_option' ) );
-	add_filter( 'pre_update_option_cron', array( \Automattic\WP\Cron_Control\Events_Store::instance(), 'update_option' ), 10, 2 );
-	add_filter( 'schedule_event', array( \Automattic\WP\Cron_Control\Events_Store::instance(), 'block_creation_if_job_exists' ) );
-	\Automattic\WP\Cron_Control\_resume_event_creation();
+
+	// Need to re-add the filters since they would have been skipped over in wp-adapter.php the first time around.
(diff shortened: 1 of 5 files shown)


tests/bootstrap.php
```diff
@@ -37,13 +37,18 @@ function _manually_load_plugin() {
...
+	// Need to re-add the filters since they would have been skipped over in wp-adapter.php the first time around.
+	add_filter( 'pre_schedule_event', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_schedule_event', 10, 2 );
+	add_filter( 'pre_reschedule_event', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_reschedule_event', 10, 2 );
+	add_filter( 'pre_unschedule_event', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_unschedule_event', 10, 4 );
+	add_filter( 'pre_clear_scheduled_hook', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_clear_scheduled_hook', 10, 3 );
+	add_filter( 'pre_unschedule_hook', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_unschedule_hook', 10, 2 );
+	add_filter( 'pre_get_scheduled_event', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_get_scheduled_event', 10, 4 );
+	add_filter( 'pre_get_ready_cron_jobs', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_get_ready_cron_jobs', 10, 1 );
+	add_filter( 'pre_option_cron', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_get_cron_option', 10 );
+	add_filter( 'pre_update_option_cron', __NAMESPACE__ . '\Automattic\WP\Cron_Control\pre_update_cron_option', 10, 2 );
```
WPprodigy, line 51: I don't love this, but it's needed atm because currently the filters aren't added unless the custom table has been installed (which makes sense, since we don't want to overtake WP's cron registration until we are ready to actually handle it).

Some other options could include to check if the table is installed inside each filter callback alongside the `null` checks. This way it would start working as soon as the table is installed, though it is a bit of overhead.

I want to look a bit deeper and more holistically into what the install process looks like a bit later.

pschoffer (approved):
Looks good. I like the rename .. find is more descriptive i think,

WPprodigy:
Yeah, I'm really liking `find()` much more as well, since it better implies that you might not find something :)

PR #226, opened 2021-12-08
Events store cleanup

WPprodigy:
"Cron control next, v2" has the base framework setup now. So time utilize the new methods for database interactions, and start deprecating unused methods (once they are truly unused by the plugin).

Now that we have a new DB interaction layer, we can quickly make the rest of the plugin "follow suit" by updating the store methods that the rest of the plugin has currently in use. This way our caching layer stays in tact and is invalidated when needed. And everything will start running again exactly as it did before.

As the usage of the old store method's are removed throughout the plugin, they will move from "deprecation coming soon" to actually deprecated w/ the `_deprecated_function()` being added so we can watch for php warnings in case of usage in the wild.

The methods that are not fully deprecated in this PR will be handled individually as other notable classes are cleaned up (really just looking at the runner/CLI for now I think). The rest will likely come post-real-merge.

### DIFF
--- a/includes/class-events-store.php
+++ b/includes/class-events-store.php
@@ -8,21 +8,13 @@
 namespace Automattic\WP\Cron_Control;
 
 /**
- * Events Store class
+ * Event's Store class
  */
 class Events_Store extends Singleton {
-	/**
-	 * PLUGIN SETUP
-	 */
-
-	/**
-	 * Class constants
-	 */
 	const TABLE_SUFFIX = 'a8c_cron_control_jobs';
 
 	const DB_VERSION        = 1;
 	const DB_VERSION_OPTION = 'a8c_cron_control_db_version';
-	const TABLE_CREATE_LOCK = 'a8c_cron_control_creating_table';
 
 	const STATUS_PENDING   = 'pending';
 	const STATUS_RUNNING   = 'running';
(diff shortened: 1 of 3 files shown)


WPprodigy:
Gonna have to increase the plugin req's to PHP 7.4...but...it's all passing 🥲

includes/class-events-store.php
```diff
@@ -999,7 +618,7 @@ public static function _query_events_raw( array $args = [] ): array {
 				'validation' => fn( $status ) => self::validate_status( $status ),
 			],
 			'limit' => [
-				'default'    => 1,
+				'default'    => 100,
 				'validation' => 'is_int',
 			],
```
WPprodigy, line 623: The only real change to the "API" is here. Seems more fitting for a method named `_query_events_raw()` to not default to just querying 1. 100 is plenty lightweight I think, and tends to be what I went w/ in various implementations already.

tests/tests/class-events-store-tests.php
```diff
@@ -148,33 +148,6 @@ function test_event_unscheduling_against_event_store() {
 		$this->assertEmpty( Utils::get_events_from_store() );
 	}
 
-	/**
-	 * Test event-cache splitting
-	 */
-	function test_excessive_event_creation() {
```
WPprodigy, line 154: There is no pre-cache-bucketing anymore, so we don't need to worry about this. I've made a note to remove this later: https://github.com/Automattic/vip-go-mu-plugins/blob/a361f213b517eb3e6a070576bb8bf34e1dafbb65/001-cron.php#L118-L126

includes/class-events-store.php
```diff
@@ -91,25 +58,14 @@ protected function class_init() {
...
-		return $wpdb->prefix . self::TABLE_SUFFIX;
-	}
-
-	/**
-	 * Set initial options that control plugin's behaviour
-	 */
-	protected function prime_options() {
-		// Prime DB option.
-		add_option( self::DB_VERSION_OPTION, 0, null, false );
+		return self::table_name();
```
pschoffer, line 68: Why this change? I am not a big fan of static functions. but I get that you now need it in the other new static functions :worried:

pschoffer (changes requested):
Looks good.. the only thing I am a bit concerned about is that all the new methods are static forcing us to move even more logic to static in this one. What was a reason for making them static again?

includes/class-events-store.php
WPprodigy: This specific change was just to help remove some of the duplication.

WPprodigy:
As far as I can tell, we don't need state in the store so static methods seemed easier for access. I don't much like this pattern: https://github.com/Automattic/Cron-Control/blob/master/includes/functions.php, makes following method usage kinda painful. And when things are static, the call signatures are already short so there's really no reason to wrapper function for them.

The overall plan was to deprecate the others methods eventually. Can switch to non-statics on everything new though if you feel that is better.

pschoffer:
I see the benefit of shorter syntax. 

The biggest issue with statics is that you can't mock them (without some pretty nasty things ) no matter how you write your code consumer.

If it is not static you can always expose a way to inject a MockedEventStore... where calling `_update_event` won't actually mess with DB.

This is the reason why I would vote for non-statics.

pschoffer (approved):
That being said.. the static methods are not actually introduced in this PR... so I should have raised that sooner or maybe we can do that after. Approving for now to not block you, but would love to get your input on previous comment

WPprodigy:
> The biggest issue with statics is that you can't mock them (without some pretty nasty things ) no matter how you write your code consumer.

That's a good point. If we ever do need "state" shared, it'll be ready for it as well. The singleton makes sense for a data store like this, and loading it up is basically free anyway. I'll address in a followup PR.

PR #227, opened 2021-12-14
Increase min versions, and add checks for the required php & wp versions

WPprodigy:
- Checks for min required wp/php version.
- Alerts when there is a problem.
- Re-organized the Main class a little.
- Updated readme/plugin header/language files.


pschoffer (approved):
LGTM

PR #228, opened 2021-12-15
Swap to non-static methods in Event's store

WPprodigy:
per feedback in https://github.com/Automattic/Cron-Control/pull/226#issuecomment-992229011

Nothing much changes, just preps making it easier in the future to inject the Store into Event/Events if we wanted to mock easier / provide a filter for swapping out w/ custom stores.

(php 7.3 tests will fail until the other PR is merged)

### DIFF
--- a/includes/class-event.php
+++ b/includes/class-event.php
@@ -131,11 +131,11 @@ class Event {
 		}
 
 		if ( $this->exists() ) {
-			$success = Events_Store::_update_event( $this->id, $row_data );
+			$success = Events_Store::instance()->_update_event( $this->id, $row_data );
 			return true === $success ? true : new WP_Error( 'cron-control:event:failed-update' );
 		}
 
-		$event_id = Events_Store::_create_event( $row_data );
+		$event_id = Events_Store::instance()->_create_event( $row_data );
 		if ( $event_id < 1 ) {
 			return new WP_Error( 'cron-control:event:failed-create' );
 		}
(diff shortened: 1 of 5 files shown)


includes/class-event.php
```diff
@@ -131,11 +131,11 @@ public function save() {
 		}
 
 		if ( $this->exists() ) {
-			$success = Events_Store::_update_event( $this->id, $row_data );
+			$success = Events_Store::instance()->_update_event( $this->id, $row_data );
```
pschoffer, line 134: This is great. One more step we could do is make $event_store a field in this class.

Then in the constructor you would do:
`this->event_store = Event_store::instance()`

And here just `this->event_store->_update_event`. 

The benefit of this is that then it is truly ready for unit testing.

In your unit test you could create an instance of Event and change the event_store to mocked one while testing these functions.

I dont think we need to rewrite this and the tests right now... but I am happy to help you out with that if you think it sounds good.

pschoffer (approved):
Awesome... left a minor note, but can be done later as well up to you sir.

includes/class-event.php
WPprodigy: Yep! Actually had something like this in the first iteration, but decided to strip it down since it kind of felt like extra code just for the sake of more code.

If/when we reach the point where we'd like to allow the store to be extensible/substitutable, or mock accordingly, then what you described can definitely be added - and in a way that won't require any BC breaks. So mainly for that reason I decided to punt it down the road for now, vs added code that maybe will, maybe won't, be used.

PR #229, opened 2021-12-16
Deprecate legacy store getters

WPprodigy:
Finishes fully deprecating the following legacy "getters" from the events store, so they are no longer used within the plugin:

```
Events_Store->get_jobs()
Events_Store->get_job_by_id()
Events_Store->get_job_by_attributes()

// And their "helpers"
get_events()
get_event_by_attributes()
get_event_by_id()
```

I tried my best to not touch too much.

### DIFF
--- a/includes/class-event.php
+++ b/includes/class-event.php
@@ -133,19 +133,29 @@ class Event {
 			$row_data['interval'] = 0;
 		}
 
+		// About to be updated, so increment the "last modified" timestamp.
+		$current_time = current_time( 'mysql', true );
+		$row_data['last_modified'] = $current_time;
+
 		if ( $this->exists() ) {
 			$success = Events_Store::instance()->_update_event( $this->id, $row_data );
-			// TODO: update last modified
-			return true === $success ? true : new WP_Error( 'cron-control:event:failed-update' );
+			if ( ! $success ) {
+				return new WP_Error( 'cron-control:event:failed-update' );
+			}
+
+			$this->last_modified = $current_time;
+			return true;
 		}
+
+		$row_data['created'] = $current_time;
 
 		$event_id = Events_Store::instance()->_create_event( $row_data );
 		if ( $event_id < 1 ) {
 			return new WP_Error( 'cron-control:event:failed-create' );
 		}
 
-		// TODO: update created/last modified
 		$this->id = $event_id;
+		$this->created = $current_time;
 		return true;
 	}
 
--- a/includes/class-event.php
+++ b/includes/class-event.php
@@ -134,12 +134,8 @@ class Event {
 		}
 
 		if ( $this->exists() ) {
-<<<<<<< HEAD
 			$success = Events_Store::instance()->_update_event( $this->id, $row_data );
-=======
-			$success = Events_Store::_update_event( $this->id, $row_data );
 			// TODO: update last modified
->>>>>>> 91f7cb0 (Cleanup Events())
 			return true === $success ? true : new WP_Error( 'cron-control:event:failed-update' );
 		}
 
--- a/includes/class-event.php
+++ b/includes/class-event.php
@@ -22,6 +22,9 @@ class Event {
 	// When the event will run next.
 	private int $timestamp;
 
+	private $created;
+	private $last_modified;
+
 	/*
 	|--------------------------------------------------------------------------
 	| Getters
@@ -131,7 +134,12 @@ class Event {
 		}
 
 		if ( $this->exists() ) {
+<<<<<<< HEAD
 			$success = Events_Store::instance()->_update_event( $this->id, $row_data );
+=======
+			$success = Events_Store::_update_event( $this->id, $row_data );
+			// TODO: update last modified
+>>>>>>> 91f7cb0 (Cleanup Events())
 			return true === $success ? true : new WP_Error( 'cron-control:event:failed-update' );
 		}
 
@@ -140,6 +148,7 @@ class Event {
 			return new WP_Error( 'cron-control:event:failed-create' );
 		}
 
+		// TODO: update created/last modified
 		$this->id = $event_id;
 		return true;
 	}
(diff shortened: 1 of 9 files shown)


includes/class-event.php
```diff
@@ -22,6 +22,9 @@ class Event {
 	// When the event will run next.
 	private int $timestamp;
 
+	private $created;
+	private $last_modified;
```
WPprodigy, line 26: Need to set these when we pull from the database for `get_legacy_event_format()` (legacy compatibility w/ hook params).

Will probably make them first-class citizens if we end up using them elsewhere.

includes/class-events.php
```diff
@@ -89,56 +91,41 @@ public function populate_concurrent_action_whitelist() {
...
 		$job_queue_window = is_null( $job_queue_window ) ? JOB_QUEUE_WINDOW_IN_SECONDS : $job_queue_window;
 
-		$events = get_option( 'cron' );
+		// Grab relevant events that are due, or soon will be.
+		$current_time = time();
+		$events = self::query( [
+			'timestamp' => [ 'from' => 0, 'to' => $current_time + $job_queue_window ],
+			'status'    => Events_Store::STATUS_PENDING,
+			'limit'     => -1, // Need to get all, to ensure we grab internals even when queue is backed up.
+		] );
```
WPprodigy, line 110: This is a pretty nice improvement :). Rather than querying ALL events, we can just grab the ones we are interested in (pending events within the time window).

includes/class-events.php
```diff
@@ -89,56 +91,41 @@ public function populate_concurrent_action_whitelist() {
...
 		foreach ( $events as $event ) {
-			// Skip events whose time hasn't come.
-			if ( $event['timestamp'] > $current_window ) {
-				continue;
-			}
-
-			// Skip events that don't have any callbacks hooked to their actions, unless their execution is requested.
-			if ( ! $this->action_has_callback_or_should_run_anyway( $event ) ) {
-				continue;
-			}
```
WPprodigy, line 129: In theory this feature makes sense, but it never worked in practice and we ended up disabling it via filter in mu-plugins b/c it would cause runaway event creation (recurring ones would just keep being recreated over and over).

So I'm just removing it altogether. I think the safer option will be to just throw a PHP warning when an event runs but has no callback.

includes/class-events.php
```diff
@@ -450,81 +360,30 @@ private function reset_event_lock( $event ) {
 	/**
 	 * Turn the event action into a string that can be used with a lock
 	 *
-	 * @param object $event Event data.
+	 * @param Event|stdClass $event
 	 * @return string
 	 */
-	public function get_lock_key_for_event_action( $event ) {
+	public function get_lock_key_for_event_action( $event ): string {
```
WPprodigy, line 366: This method is public, and currently used in the lock class, so have to keep BC for now.

In the future I think either locks because first-class citizens of the Event object, (i.e. `$event->get_lock_key()`), or they are going to get a rework / maybe removed altogether. So just leaving more or less as is for now.

includes/class-events.php
```diff
@@ -450,81 +360,30 @@ private function reset_event_lock( $event ) {
...
 		// Hashed solely to constrain overall length.
-		return md5( 'ev-' . $event->action );
-	}
-
-	/**
-	 * Mark an event completed, and reschedule when requested
-	 *
-	 * @param object $event Event data.
-	 */
-	private function update_event_record( $event ) {
```
WPprodigy, line 466: `$event->reschedule();` already encapsulates all this logic, so we can just drop it :)

includes/wp-cli/class-events.php
```diff
@@ -175,7 +179,7 @@ public function run_event( $args, $assoc_args ) {
 	 * @param array $assoc_args Array of flags.
 	 * @return array
 	 */
-	private function get_events( $args, $assoc_args ) {
+	private function retrieve_events( $args, $assoc_args ) {
```
WPprodigy, line 182: `get_events()` as a method named is used differently throughout the plugin, so I wanted to differentiate this some to make my code-searching life easier since it's private.

includes/wp-cli/class-rest-api.php
```diff
@@ -86,24 +88,22 @@ public function get_queue( $args, $assoc_args ) {
 	private function format_events( $events ) {
```
WPprodigy, line 88: This is...a bit weird. It's a CLI class that runs the REST endpoints. Honestly thinking we can probably 'nix this whole part later on.

includes/class-events.php
```diff
@@ -89,56 +91,41 @@ public function populate_concurrent_action_whitelist() {
...
 	 *
-	 * @param array $job_queue_size   Maximum number of events to return (excludes internal events).
-	 * @param array $job_queue_window How many seconds into the future events should be fetched.
-	 * @return array
+	 * @param null|int $job_queue_size   Maximum number of events to return (excludes internal events).
+	 * @param null|int $job_queue_window How many seconds into the future events should be fetched.
+	 * @return array Events to be run in the next batch.
 	 */
-	public function get_events( $job_queue_size = null, $job_queue_window = null ) {
+	public function get_events( $job_queue_size = null, $job_queue_window = null ): array {
```
WPprodigy, line 100: This is unfortunately named per what it does. As `Events/get_events()`, it actually more specifically fetches "due" events for the next round of cron running. Would be more aptly named `get_next_due_batch()` or something, so probably going to aim for some deprecation here in the future.

includes/class-event.php
```diff
@@ -132,6 +135,7 @@ public function save() {
 
 		if ( $this->exists() ) {
 			$success = Events_Store::_update_event( $this->id, $row_data );
+			// TODO: update last modified
```
pschoffer, line 138: Are you going to implement these TODOs?

WPprodigy: Was thinking to address is separately. There's not really any harm in them "being stale" atm, but if ever exposing them w/ their own getters or logic, then I'd want the class to keep them fresh.

Probably means setting the dates within save() and passing it to the store methods, rather than setting them within the store.

WPprodigy:
(rebased)

includes/class-event.php
WPprodigy: Okay turns out this was much easier than I thought since everything is funneled through the same flow now!

Both TODOs account for :)

WPprodigy:
Gonna merge this to prevent some annoying conflicts while working on deprecating the legacy store "setters". But please do still leave any feedback and I'll address in follow-up commits.

PR #230, opened 2021-12-17
Deprecate legacy store setters

WPprodigy:
Finished fully deprecating the below methods from the event's store. Interesting to note that I pretty much only had to touch a few CLIs for this final round, and none of the orchestrator clis at that. So overall impact/risk here is quite low, though yes I have tested and will include final testing for these as well in the last PR when we go to `master`.

```
create_or_update_job()
mark_job_completed()
mark_job_record_completed()

flush_internal_caches()
suspend_event_creation()
resume_event_creation()

// And their associated "helper" functions.
schedule_event()
delete_event()
delete_event_by_id()
_flush_internal_caches()
_suspend_event_creation()
_resume_event_creation()
```

Also removes the following CLIs (explanation in comments to follow):

```
wp cron-control cache flush
wp cron-control-fixers remove-all-plugin-data
wp cron-control-fixers purge-legacy-cpt-entries
```

___

The only legacy "event related" store methods remaining are now just `purge_completed_events()` & `count_events_by_status()`. I have plans for reworking the completed event purging in the future, but leaving that alone for now. As for counting events, probably in the future will have `_query_events_raw()` have an arg for counting instead of returning rows, then `Events::query_count( $args )` as a wrapper for getting the count. But can do that later as well if we find the counting is still needed (after general CLI cleanup).


includes/wp-cli/class-cache.php
```diff
@@ -1,32 +0,0 @@
...
-	/**
-	 * Flush the cache
-	 *
-	 * @subcommand flush
-	 * @param array $args Array of positional arguments.
-	 * @param array $assoc_args Array of flags.
-	 */
-	public function flush_internal_caches( $args, $assoc_args ) {
-		$flushed = \Automattic\WP\Cron_Control\_flush_internal_caches();
-
```
WPprodigy, line 23: I want to stop exposing this from the store. At first was going to do a `WP_CLI::error( "deprecated msg..." )`, but realized you get basically the same result by removing the CLI entirely. WP just shows like `Error: 'X' is not a registered wp command.`, which feels more than sufficient.

Removing in general also feels a lot safer in CLI context.

includes/wp-cli/class-cache.php
```diff
@@ -1,32 +0,0 @@
...
-class Cache extends \WP_CLI_Command {
-	/**
-	 * Flush the cache
-	 *
-	 * @subcommand flush
-	 * @param array $args Array of positional arguments.
-	 * @param array $assoc_args Array of flags.
-	 */
-	public function flush_internal_caches( $args, $assoc_args ) {
-		$flushed = \Automattic\WP\Cron_Control\_flush_internal_caches();
```
WPprodigy, line 22: I want to stop exposing this detail from the store, as it can be handled solely from within the store anytime DB interactions take place.

At first was going to do a `WP_CLI::error( "deprecated msg..." )`, but realized you get basically the same result by removing the CLI entirely. WP just shows like "Error: 'X' is not a registered wp command.", which feels more than sufficient.

Removing in general also feels a lot safer in CLI context.

includes/wp-cli/class-events.php
```diff
@@ -434,14 +432,9 @@ private function delete_event_by_id( $args, $assoc_args ) {
...
-	 * Delete all events of the same action
-	 *
-	 * @param array $args Array of positional arguments.
-	 * @param array $assoc_args Array of flags.
+	 * Delete all events of the same action.
 	 */
-	private function delete_event_by_action( $args, $assoc_args ) {
-		$action = $assoc_args['action'];
-
+	private function delete_event_by_action( $action ) {
```
WPprodigy, line 437: Rather greatly just simplified this CLI method. Functionally it works the same, just less verbose. I tested w/ a 25k event deletion, and all went well. There is room to optimize if needed, but this is pretty rarely-used functionality, so I'm not really worried.

includes/wp-cli/class-one-time-fixers.php
```diff
@@ -1,179 +0,0 @@
...
-			\WP_CLI::confirm( __( 'Proceed?', 'automattic-cron-control' ) );
-			\WP_CLI::log( "\n" . __( 'Starting...', 'automattic-cron-control' ) . "\n" );
-		}
-
-		// Don't create new events while deleting events.
-		\Automattic\WP\Cron_Control\_suspend_event_creation();
-
-		// Don't truncate as it requires DROP and resets auto-increment value.
-		if ( ! $dry_run ) {
-			$wpdb->query( "DELETE FROM {$table_name}" ); // Cannot prepare table name. @codingStandardsIgnoreLine
```
WPprodigy, line 61: Started by moving this to a new store method, but ultimately decided it's not really something we need. If somebody is embarking w/ a goal like this, they'll have to know what they are doing and can pretty much just as easily run the delete SQL manually. At least just as easily as one could track down this CLI, if not more likely :)

And similar to the cache CLI, rather than log a CLI error, can just delete the whole CLI pretty safely.

includes/wp-cli/class-one-time-fixers.php
```diff
@@ -1,179 +0,0 @@
...
-
-	/**
-	 * Remove Cron Control data previously stored in a CPT
-	 *
-	 * @subcommand purge-legacy-cpt-entries
-	 * @synopsis [--batch-size=<batch-size>] [--dry-run=<dry-run>]
-	 * @param array $args Array of positional arguments.
-	 * @param array $assoc_args Array of flags.
-	 */
-	public function purge_cpt( $args, $assoc_args ) {
```
WPprodigy, line 86: Since we're dropping the `remove-all-plugin-data` CLI, this was the last one left in the file and well, it's even less needed. It's the last relic in the plugin from when CPTs were used. It has been quite a few years, safe to drop it I think.

pschoffer (approved):
Great stuff +115 −587 \o/

WPprodigy:
Wait until we can actually remove the unused methods 😲

rinatkhaziev:
@WPprodigy did we check mu-plugins for the usage of any of those? I think there may be one or two cases

WPprodigy:
Not seeing anything on a quick look, usually we still use WP core functions, else just get the custom table name and do our querying through that right now (which I'll look to change later).

PR #235, opened 2021-12-22
Loosen the "timestamp" int type enforcement

WPprodigy:
The "run event" CLIs via the orchestrator are failing when they can't find the event in question b/c the CLI is passing around strings for timestamps rather than ints.

Upon further looking, core docblocks also state it will always be an `int` passed for timestamps like in `wp_get_scheduled_event()`, but there is really no enforcement from what the user passes in and what gets sent onto the filters. So for proper compatibility, we should swap to using `is_numeric()` b/c that is how WP core treats the timestamps throughout its logic.

Also increasing the query error to a Warning so we'll catch things like this in logstash.


pschoffer (approved):
Makes sense!

PR #236, opened 2021-12-23
Implement extra caching layer for events

WPprodigy:
Kind of put this off initially due to extra complexity and not wanting to pre-optimize w/o data, but thinking it's likely going to be desired. So starting the work on it ahead of time.

In short, this is perhaps the most common pattern on the frontend:

```
if ( ! wp_next_scheduled ( 'action_name', $args ) ) {
  wp_schedule_event( time(), 'hourly', 'action_name', $args );
}
```

Right now we flush the whole query cache anytime an event is updated (such as when events are run). This means the next FE request will need to do a fresh query for that `wp_next_scheduled()` call, which can amount to quite a number throughout the codebase.

This PR addresses this by adding an additional caching layer that we can more selectively invalidate. As events are run, the FE shouldn't see much impact under normal conditions.

### DIFF
--- a/includes/class-event.php
+++ b/includes/class-event.php
@@ -177,7 +177,7 @@ class Event {
 		// Prevent conflicts with the unique constraints in the table.
 		// Is a bit unfortunate since it won't be as easy to query for the event anymore.
 		// Perhaps in the future could remove the unique constraint in favor of stricter duplicate checking.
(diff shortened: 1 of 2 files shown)


pschoffer (approved):
Took me a while. But looks great:) I am not sure how to make it easier to read so :ship:

Issue #159, opened 2018-03-27
Block events with excessively large arguments


nickdaugherty:
For stability, cron should not accept events with arguments of excessive size, because this causes problems with caching, db load, network transfer, all sorts of fun stuff.

Usually these types of cron events are storing full objects as arguments, instead of ids - like terms for example.

There should never be a situation where a cron event contains 400kb of data 😱 

Need to find the right cutoff, but 10k might be a good starting point?

nickdaugherty:
Need to think about how to surface failures - can't silently fail.

ethitter:
> Need to think about how to surface failures - can't silently fail.

Indeed. Not scheduling the event will essentially create a race condition as the code continually tries to do so, as we saw with the "skip events with no callbacks" feature.

To avoid that, it might make sense to schedule the event but drop the arguments, replacing them with some flag to delete the event without running. As long as the action, timestamp, and instance are properly set, Cron Control shouldn't keep rescheduling a bad event.

nickdaugherty:
Good catch - we should definitely do that to prevent repeated re-scheduling.

WPprodigy:
This should be a pretty non-issue when we can stop relying on the `cron` option completely. Once the new core filters are implemented, almost all use cases for `cron` are gone (except backwards compatibility w/ custom plugin code).

There are only two places WP core still uses `_get_cron_array()` w/o it being filterable beforehand, but these are rare and not super mission critical (site health page, media table reading a `upgrader_scheduled_cleanup` job).

tl;dr, we can "soon" make the `cron` option not auto-loaded. It can still return the combined bucketed data when it's called manually (and perhaps we throw deprecating notice/warning)

WPprodigy:
W/ the upcoming changes, this should mostly be a non-issue. If custom code is still calling `get_option('cron')`, at most we cache in "buckets" of 100 events.

I'm thinking we can start w/ just throwing a PHP warning during event (re)scheduling if the args is rather significantly large, that way it'll be notable when debugging application performance concerns. But we won't run into any troubles by deviating from expected functionality.

WPprodigy:
Started logging for this here: https://github.com/Automattic/Cron-Control/pull/237. Pretty gracious on the limit, decided on something that will enable to cache up to 500 events per cache key. Seems reasonable.

PR #237, opened 2021-12-24
Check args size, and increase query limit

WPprodigy:
Addresses https://github.com/Automattic/Cron-Control/issues/159. Instead of blocking events w/ large arguments though (which could cause other problems), just warn for now so it is at least visible.

W/ this new checking in place, it's safer to go ahead and increase how many events we query at a time under the presumption that events will stay under this "reasonable limit". This really only matters if there is still code doing `get_option( 'cron' )` somewhere, as that is when we need to retrieve all events.

Increasing from 100 to 500 is a pretty good improvement, as a site w/ say 3000 active events will need to do just 6 queries instead of 30. Of course they are cached for subsequent requests (assuming args stay under the, quite gracious, size limit), but the cache is invalidated whenever events are updated/run - so it will still take place fairly often should code like the `get_option( 'cron' )` still exist on the frontend.

Worst case, if a ton of events w/ massive args are added, the query will fail to save some result sets in memcache and instead will need to re-query each time. It's not the end of the world since the queries for "all events" are quite performant since we are talking about a pretty small table, and the warnings should be noticeable.
 
_____

I ran quite a few size tests, writing them down here for posterity. Could probably go higher than expecting to cache just 500 events per cache key, but seems like a reasonable and safe number for now.

Test event creation:

```
add_action( 'init', function() {
  for ( $i = 0; $i <= 1000; $i++ ) {
    // no args
    $args = [];
    wp_schedule_event( time(), 'hourly', 'some_action1', [$i, $args] );

    // w/ medium sized args
    $args = array_fill( 0, 100, 'a' );
    wp_schedule_event( time(), 'hourly', 'some_action2', [$i, $args] );

    // w/ big args
    $args = array_fill( 0, 100, str_repeat( 'a', 100 ) );
    wp_schedule_event( time(), 'hourly', 'some_action3', [$i, $args] );

    // w/ bigger array, smaller content
    $args = array_fill( 0, 1000, 'a' );
    wp_schedule_event( time(), 'hourly', 'some_action4', [$i, $args] );

    // quite massive
    $args = array_fill( 0, 1000, str_repeat( 'a', 1000 ) );
    wp_schedule_event( time(), 'hourly', 'some_action5', [$i, $args] );
  }
} );
```

Sizing checks:

```
// No args:

$test = Automattic\WP\Cron_Control\Events_Store::instance()->_query_events_raw( [ 'limit' => 100, 'page' => 1, 'action' => 'some_action1' ] );
mb_strlen( serialize( $test ), '8bit' )
=> int(44068) // .044mb

____

// w/ medium sized args
$args = array_fill( 0, 100, 'a' );

$test = Automattic\WP\Cron_Control\Events_Store::instance()->_query_events_raw( [ 'limit' => 100, 'page' => 1, 'action' => 'some_action2' ] );
mb_strlen( serialize( $test ), '8bit' )
=> int(173609) // 0.17mb

____

// W/ big args
$args = array_fill( 0, 1000, str_repeat( 'a', 100 ) ) ;

$test = Automattic\WP\Cron_Control\Events_Store::instance()->_query_events_raw( [ 'limit' => 100, 'page' => 1, 'action' => 'some_action3' ] );
mb_strlen( serialize( $test ), '8bit' )
=> int(1183893) // 1.1mb

// compressed
mb_strlen( gzdeflate( serialize( $test ) ), '8bit' )
=> int(16050) .01mb
____

// W/ bigger array, smaller content
$args = array_fill( 0, 1000, 'a' ) ;

$test = Automattic\WP\Cron_Control\Events_Store::instance()->_query_events_raw( [ 'limit' => 100, 'page' => 1, 'action' => 'some_action4' ] );
mb_strlen( serialize( $test ), '8bit' )
=> int(1433672) // 1.4mb

// compressed
strlen( gzdeflate( serialize( $test ) ) );
mb_strlen( gzdeflate( serialize( $test ) ), '8bit' )
=> int(28508) // .02mb

____

Quite Massive

$args = array_fill( 0, 1000, str_repeat( 'a', 1000 ) ) ;

$test = Automattic\WP\Cron_Control\Events_Store::instance()->_query_events_raw( [ 'limit' => 100, 'page' => 1, 'action' => 'some_action5' ] );
mb_strlen( serialize( $test ), '8bit' )
=> int(11533771) // 11.5mb

// compressed
mb_strlen( gzdeflate( serialize( $test ) ), '8bit' )
=> int(309113) .3mb
```

### DIFF
--- a/includes/class-event.php
+++ b/includes/class-event.php
@@ -368,8 +368,7 @@ class Event {
 		$db_row_size_of_normal_event = 300;
 
 		// Note: Memcache can only cache up to 1mb values, after compression.
-		$default_storage_needed_for_events = $db_row_size_of_normal_event * $max_events_per_page;
-		$reasonable_size = ( MB_IN_BYTES - $default_storage_needed_for_events ) / $max_events_per_page;
+		$reasonable_size = ( MB_IN_BYTES / $max_events_per_page ) - $db_row_size_of_normal_event;
 
 		// First a quick uncompressed test.
 		$uncompressed_size = mb_strlen( serialize( $this->get_args() ), '8bit' );
--- a/includes/class-event.php
+++ b/includes/class-event.php
@@ -325,6 +325,11 @@ class Event {
 			}
 		}
 
+		// Don't prevent event creation, but do warn about overly large arguments.
+		if ( ! $this->args_array_is_reasonably_sized() ) {
+			trigger_error( sprintf( 'Cron-Control: Event (action: %s) was added w/ abnormally large arguments. This can badly effect performance.', $this->get_action() ), E_USER_WARNING );
+		}
+
 		return true;
 	}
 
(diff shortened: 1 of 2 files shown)


includes/class-event.php
```diff
@@ -354,4 +359,33 @@ private function get_refreshed_schedule_interval() {
...
+	private function args_array_is_reasonably_sized(): bool {
+		// We aim to cache queries w/ up to 500 events.
+		$max_events_per_page = 500;
+
+		// A compressed db row of an event is around 300 bytes.
+		$db_row_size_of_normal_event = 300;
+
+		// Note: Memcache can only cache up to 1mb values, after compression.
+		$default_storage_needed_for_events = $db_row_size_of_normal_event * $max_events_per_page;
+		$reasonable_size = ( MB_IN_BYTES - $default_storage_needed_for_events ) / $max_events_per_page;
```
pschoffer, line 372: Would this simplification still get you the size you want?

```suggestion
		// A compressed db row of an event is around 300 bytes.
		$db_row_size_of_normal_event = 300;

		// Note: Memcache can only cache up to 1mb values, after compression.
		$reasonable_size = ( MB_IN_BYTES / $max_events_per_page ) - $db_row_size_of_normal_event;
```
Is $db_row_size_of_normal_event kinda the amount you need to store the stuff around the event excluding the actual arguments?

pschoffer (changes requested):
I have one question/not sure I understand one part. Otherwise looks good

includes/class-event.php
WPprodigy: Right, it's roughly the size of a normal event row w/ an empty args array, after compression. I rounded up a little for some extra padding.

PR #244, opened 2022-01-12
Validate schedule_event filter

WPprodigy:
WP core kinda set a precedent for this filter being used as a mechanism for preventing events to be scheduled, as various plugins will return `false`. So safeguarding against that and any other filter weirdness.

Currently seeing fatals like this as a result:

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Automattic\WP\Cron_Control\Event::set_action() must be of the type string, null given, called in /var/www/wp-content/mu-plugins/cron-control/includes/wp-adapter.php on line 70 and defined in /var/www/wp-content/mu-plugins/cron-control/includes/class-event.php:77
```

### DIFF
--- a/includes/wp-adapter.php
+++ b/includes/wp-adapter.php
@@ -64,6 +64,9 @@ function pre_schedule_event( $pre, $event ) {
 
 	/** This filter is documented in wordpress/wp-includes/cron.php */
 	$event = apply_filters( 'schedule_event', $event );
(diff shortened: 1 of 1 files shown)


rinatkhaziev (approved):

WPprodigy:
Oh meant to link this for context: https://github.com/WordPress/WordPress/blob/92f0b0cb48a3de660655cc658fc739695a393096/wp-includes/cron.php#L165-L192

Issue #232, opened 2021-12-17
Task: Cleanup database installation process


WPprodigy:
## 1) Ensure table truly exists

Problem: Somebody may duplicate the options table while creating a new subsite locally for example, but for some reason not copy over the cron control table. As a result, the `a8c_cron_control_db_version` option is set but the table doesn't exist - and will never be attempted to be created. So you end up with perpetual errors until manually resolved.

Solution: Not 100% sure, kind of tricky. This would be an ideal for a cron job to verify the table exists, maybe in an existing internal cleanup job. But uhh, cron is broken if the table isn't there. So maybe need to do a SQL query to just check that the table truly exists every once in a while? Alternatively, try to leave cron in a functional state when this happens - but that may be hard.

____

## 2) Cleanup overall table creation flows

There are lots of different flows for when/how a table is created, might be good to simplify a bit where possible. Of note, `wpmu_new_blog` is a deprecated hook that needs to be updated a minimum.

____

## 3) Prepare for schema changes

The plugin won't currently react very well to just incrementing the db_version option if we ever do schema updates, as it won't actually update itself.

Notably eying changing the `ts_action_instance_status` index to be re-ordered to `action` -> `instance` -> `timestamp` -> `status` instead allowing it to be used for action-only queries as well (which WP core has a few of).

WPprodigy:
Items 1 & 2 from above are resolved in https://github.com/Automattic/Cron-Control/pull/245.

Supporting schema changes can be dealt with when the time comes for them.

WPprodigy:
New item fresh off the press: we need to actively set the charset for our table upon creation, like this: https://github.com/WordPress/WordPress/blob/69e9e78cf4b63931a3d81e7ecb3c2a0cda7a13ed/wp-admin/includes/schema.php#L64

PR #245, opened 2022-01-17
Update table install process

WPprodigy:
- Helps resolve an issue noted in https://github.com/Automattic/Cron-Control/issues/232 where a site has the option set but doesn't actually have the table installed. So instead the `is_installed()` check now directly checks if the table exists and use the site cache instead of relying just on the option.
- `wpmu_new_blog` is deprecated, so swapping it out for `wp_insert_site`.
- Removes some of the direct side-effects from file-loading, and instead more decisively bootstraps the plugin.

Also simplified the usage a bit, now just have the public methods listed below. I checked the methods I removed and didn't see any usage, but because they were public I've added them back with deprecation notices just for extra safety during updates.

```
is_installed()
install()
maybe_install_during_shutdown()
drop_tables_on_subsite_removal()
```

The only piece this PR doesn't solve for is the "Prepare for schema changes" action item. Will handle that separately when the time comes, probably with `needs_db_update()` and `maybe_run_database_update()` methods.

## Testing

Single Site:
1) Setup a fresh single site and check that the table is created.
2) Delete the table manually and flush cache. Reload and see if table is created.

Multisite:
1) Setup a fresh multisite and check that the table is created on the root site.
2) Create a new subsite and check the the table is created.
3) Delete the table on the root site and subsite, flush cache on both. Reload and ensure table was added back.

### DIFF
--- a/includes/wp-adapter.php
+++ b/includes/wp-adapter.php
@@ -5,8 +5,7 @@ namespace Automattic\WP\Cron_Control;
 use Automattic\WP\Cron_Control\Events_Store;
 use WP_Error;
 
-// Integrate w/ WP's cron api once the custom table is installed.
-if ( Events_Store::is_installed() ) {
+function register_adapter_hooks() {
 	// Core filters added in WP 5.1, allowing us to fully use the custom event's store.
 	add_filter( 'pre_schedule_event', __NAMESPACE__ . '\\pre_schedule_event', 10, 2 );
 	add_filter( 'pre_reschedule_event', __NAMESPACE__ . '\\pre_reschedule_event', 10, 2 );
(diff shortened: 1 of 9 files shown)


includes/wp-adapter.php
```diff
@@ -303,7 +302,7 @@ function pre_update_cron_option( $new_value, $old_value ) {
 			'args'      => $event_to_add['args'],
 		];
 
-		if ( ! empty( $item['value']['schedule'] ) ) {
+		if ( ! empty( $event_to_add['schedule'] ) ) {
```
WPprodigy, line 305: Also including small fix here. The bug would have been that if somebody called `update_option( 'cron'...` directly and had added new recurring cron jobs, we'll currently end up registering them as one-time jobs instead.

Pretty rare scenario, as this is like the backup-backwards-compatibility later, but yeah good to fix for sure :)

WPprodigy: Gonna pop this into it's own PR w/ tests actually.

WPprodigy:
Did some pretty extensive testing, all looks good.

I like that the install process is rather deterministic. Before it's had the chance to install the custom table, it will carry on and use the the core `cron` option per a default wp installation. It will install the table on either the installation request (for new subsites for example) or on `shutdown` of an admin/cli/cron request - but it carries on using the `cron` option per usual. Only on the next request does it start using the new tables and our custom functionality - so there is never a scenario where it'll half use `cron` option and half use the custom table.

rebeccahum (approved):

PR #248, opened 2022-01-19
Cleanup unit tests

WPprodigy:
- Move all test-related files into `__tests__`, and adjust the internal folder structure and file naming conventions.
- Cleanup the unit tests themselves, re-using common utils and using the `Event` objects more commonly.
- Deprecate `collapse_events_array()` and `inflate_collapsed_events_array` as they are no longer used in the plugin.


pschoffer (approved):
All tests are still passing so, great!

WPprodigy:
(rebased), and added two fixes that cleanup the phpunit output and get rid of annoying php notices.

Issue #203, opened 2021-04-01
Migrated cron events may never be registered


WPprodigy:
### Use case example

WC registers some cron events during the install process: https://github.com/woocommerce/woocommerce/blob/23710744c01ded649d6a94a4eaea8745e543159f/includes/class-wc-install.php#L512. So on a site using default wp cron, the events would be registered in the `cron` option.

If the site then migrates to a new server that has cron control enabled though, those cron events do not come with. Cron-control currently uses the `pre_option_cron` to return what is in it's cron storage (not what's in the cron option). And the `pre_update_option_cron` passes through what that first filter returned to it. So after the first save to `cron` after the migration, the WC events are deleted away.

The same could apply to sites not migrating servers but just installing cron control for the first time.

### Suggestion

Cron control could check the `cron` option, and if it finds anything not registered yet it will add it to it's own storage mechanic. Particularly after https://github.com/Automattic/Cron-Control/pull/178, the `cron` option will be useless with cron control in use. So zero-ing it out would be beneficial if for nothing else than to lower the amount of data coming form the db / memcache.

Perhaps need some safety measures though. But at a minimum, need to have some method to do this at least one-time for "new" sites?

WPprodigy:
Alternatively, could also prepare for the use case of moving off of cron-control (putting events back into the `cron` option). So perhaps there could be one-time CLIs for migration purposes?

```
wp cron-control migration install
wp cron-control migration uninstall
```

ethitter:
> WC registers some cron events during the install process

I don't think WP expects cron to be used this way, so I'm not sure this problem is one to be addressed in Cron Control. The plugin should be registering its jobs more often, using Core's helpers to check if the job needs to be scheduled. It may not be necessary to register the events on `init`, since that would happen for front-end requests, but an `admin_init` callback that either registers the jobs, or hooks something to `shutdown` to do so, is more correct than registering once and hoping the jobs persist. In my experience, there are a number of reasons one may have to delete the `cron` option, with the understanding that the active code will re-register its jobs. 

> Particularly after #178, the cron option will be useless with cron control in use. So zero-ing it out would be beneficial if for nothing else than to lower the amount of data coming from the db / memcache.

The `cron` option is never used when Cron Control is enabled. #178 changes how we intercept cron events, but doesn't change the storage mechanism (custom table).

[...]

PR #250, opened 2022-01-20
Migrate legacy cron events from the cron option

WPprodigy:
Cleaned up the internal events class some in the process, but there are just two main bits of functionality changes here:

1) Internal events will only be registered after the custom table has been installed. Does not really make sense to have them before that.
2) The `clean_legacy_data()` daily cron job will now migrate events from the `cron` option to our custom table, if they don't exist still.

This is mainly an improvement for sites that are moved onto cron control after already existing for a while. It fixes https://github.com/Automattic/Cron-Control/issues/203, where notably WooCommerce cron jobs will go missing after a site is moved onto cron control. But also helps cover a range of other cases where recurring events are conditionally registered (like during plugin install), or where specific single events exist for future tasks but would currently be lost forever if a site migrated to cron control before the event ran.

Also provides a tiny bit of extra "compatibility", should some plugin decide to manually add their events directly to the `cron` option via SQL.

## Testing

The unit test is pretty good for this :)

I've also done a fresh WP install and confirmed the internal events are no longer added to the `cron` option but instead go into the table once it's been created.

And then did another fresh install (without cron control), registered a custom event manually, then activated cron cron. After running cron jobs `wp cron event run --due-now`, the custom event was moved to the custom table

### DIFF
--- a/__tests__/unit-tests/test-internal-events.php
+++ b/__tests__/unit-tests/test-internal-events.php
@@ -28,5 +28,37 @@ class Internal_Events_Tests extends \WP_UnitTestCase {
 		}
 	}
 
-	// TODO: Test the actual logic that internal events run.
+	function test_migrate_legacy_cron_events() {
+		global $wpdb;
+
+		// Ensure we start with an empty cron option.
+		delete_option( 'cron' );
+
+		// Create one saved event and two unsaved events.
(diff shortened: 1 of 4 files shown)


pschoffer (approved):
Looks good. Great job with the tests!

Issue #234, opened 2021-12-17
Task: Duplicate event prevention/cleanup


WPprodigy:
Problem: There are apparently a variety of situations where you can end up w/ duplicate events, sometimes hundreds to thousands of them. This can pretty much perpetually bog down a site until cleaned up manually.

## 1) Prevent duplicate recurring events

I actually think this is something WP core overlooked, and should probably revisit in core trac. Single-events are prevented from being registered if there is a similar action/args event already scheduled within +/-10mins. Docs actually [say the same](https://developer.wordpress.org/reference/functions/wp_schedule_event/) about recurring events, but it's not true.

Part of this is already implemented in an upcoming release, seen here: https://github.com/Automattic/Cron-Control/pull/231/commits/bbb53aeb9d83cd753f9c584b2f38c3ce91a3992c#diff-467459221a72437e5497c04a9ac91579c53ef140b9da6bb0b67c043dc1fb0490R41-R61. If an event has the same action/args/schedule, then we're just gonna deviate from core a little and prevent registration. If multiple of the same event on the same schedule is needed for some reason, then you'd just need to add some varying args when registering them.

Perhaps could take this a step further and either re-query with SRTM (send reads to master), or have a stricter unique constraint on the table (though could be tricky w/ varying event statuses).

## 2) Cleanup existing duplicate recurring events

There are quite a lot of sites in the wild now with duplicate recurring events, some to extreme degrees with hundreds of each. On a large MS, this means cron is going to struggle unnecessarily, and just have an overall heavy load that we can prevent.

So w/ the new contract in place where we prevent duplicate recurring events, we can then go through and cleanup any existing duplicates, as well as ones that may continue to slip through the cracks due to other race conditions. I'm thinking the daily `a8c_cron_control_clean_legacy_data` internal job would be great for this.

___

We'll need to document this "core deviance" in the readme.

WPprodigy:
Cleaning up duplicate recurring events now on the daily cronjob, so that's good.

Next steps I'm thinking:

1) When we determine an event doesn't exist yet (performantly), do another query that locks the table while it does an additional check and then inserts it's event. This way if there are two truly parallel requests, only one will win and get the write: https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html

[...]

PR #251, opened 2022-01-20
Prune duplicate recurring events

WPprodigy:
As apart of https://github.com/Automattic/Cron-Control/issues/234, this will help cleanup duplicate events that exist (daily). It's pretty safe, as we only look at recurring events that have the exact same action/args/schedule. There is really no need to have events like that. If you need the same hook on the same schedule, then the args need to be different to make them unique. If the args don't matter though, then probably the event should be looked into being added onto a schedule that occurs more often.

This alone won't solve all our problems mentioned in that issue, but no matter what we're going to need to cleanup the messes that currently exist. So this seems like a good starting point. Notably, this should drastically help reduce the "duplicate entry" db errors that can cause events to become stuck at the top of the queue and repeatedly run each cycle.

I opted for not including duplicate single events, since they technically clean themselves up. But could possibly revisit that in the future. For those will probably want to respect the "10minute window" that core checks: https://developer.wordpress.org/reference/functions/wp_schedule_single_event/. As in, only remove duplicate single events that exist within 10minutes of each other.

## Testing
The unit test is pretty good for this  as well :)

Did fresh WP install w/ cron control disabled and ran the following:

```
wp shell

wp_schedule_single_event( time() + 100, 'custom_single_event' );
wp_schedule_single_event( time() + 200, 'custom_single_event' );

wp_schedule_event( time() + 100, 'hourly', 'custom_recurring_event' );
wp_schedule_event( time() + 200, 'hourly', 'custom_recurring_event' );
```

Then activated cron control, and refreshed twice so the table & internal events could be setup. Then ran cron (so the cleanup legacy data task runs). Afterwards, both `custom_single_event` were in the custom table and only one `custom_recurring_event` was left.

### DIFF
--- a/includes/class-internal-events.php
+++ b/includes/class-internal-events.php
@@ -216,6 +216,7 @@ class Internal_Events extends Singleton {
 		// While this plugin doesn't use this locking mechanism, other code may check the value.
 		delete_transient( 'doing_cron' );
 
+		$this->prune_duplicate_events();
 		$this->ensure_internal_events_have_correct_schedule();
 	}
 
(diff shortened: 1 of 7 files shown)


includes/class-internal-events.php
```diff
@@ -255,6 +256,34 @@ private function migrate_legacy_cron_events() {
...
 
+	// Recurring events that have the same action/args/schedule are unnecessary. We can safely remove them.
+	private function prune_duplicate_events() {
+		$events = Events::query( [ 'limit' => -1, 'orderby' => 'ID', 'order' => 'ASC' ] );
+
+		$original_events  = [];
+		$duplicate_events = [];
+		foreach ( $events as $event ) {
+			if ( ! $event->is_recurring() ) {
+				// Only interested in recurring events.
```
rebeccahum, line 267: Is this because we may have single scheduled events with the same args that are supposed to run synchronously?

WPprodigy: Sorta. Wrote some about it in the description:

> I opted for not including duplicate single events, since they technically clean themselves up. But could possibly revisit that in the future. For those will probably want to respect the "10minute window" that core checks: https://developer.wordpress.org/reference/functions/wp_schedule_single_event/. As in, only remove duplicate single events that exist within 10minutes of each other.

rebeccahum (approved):

Issue #323, opened 2022-12-21
PHP8: Deprecated: trim(): Passing null to parameter #1


trepmal:
Notice:

```
PHP message: Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/wp-content/mu-plugins/cron-control/includes/utils.php on line 130
```

https://github.com/Automattic/Cron-Control/blob/73b9aa2722d238bdd48866be131ab0b3c5d6310b/includes/utils.php#L130

```
wp> parse_url( 'https://domain.tld', PHP_URL_PATH )
NULL
```

The [WP Core version of this function](https://github.com/WordPress/WordPress/commit/5ca2a817abb025a01bf9a8056ed00c766f7e6eed) has been updated to resolve this problem.

PR #331, opened 2023-01-16
Fix php deprecation notice with trim()

WPprodigy:
Fixes https://github.com/Automattic/Cron-Control/issues/323

Uses the same fix applied in core here: https://github.com/WordPress/WordPress/commit/5ca2a817abb025a01bf9a8056ed00c766f7e6eed

### Testing

With a test environment on php 8.1:

```
POST http://php81.vipdev.lndo.site/wp-json/cron-control/v1/events

{
  "secret": "this-is-a-secret"
}
```

You'll get a php error before the patch, and none after (but the endpoint still works of course).

Notably, the site's home_url() needs to have no subpath.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -127,27 +127,36 @@ function parse_request() {
 
 	list( $req_uri ) = explode( '?', $_SERVER['REQUEST_URI'] );
 	$self = $_SERVER['PHP_SELF'];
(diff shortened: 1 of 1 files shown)


ethitter (approved):

WPprodigy:
(fixing up tests in another PR)

PR #332, opened 2023-01-16
Update dependencies & fix tests

WPprodigy:
- Composer updated (and a few test fixups for compat w/ newer versions)
- NPM updated
- GH actions updated
- Use node 18 instead of 14


WPprodigy:
Also switched the phpcs gh action to install composer packages and use however the project is setup there (versus having to re-declare packages inside the GH action)

rebeccahum (approved):

PR #348, opened 2023-03-19
Add the package type to composer.json

rafaucau:
This Pull Request adds a package type to `composer.json`.
When managing WordPress installation via [Composer](https://github.com/roots/bedrock/blob/master/composer.json#L61), it allows installing the plugin using the `composer require automattic/cron-control` command. Without the package type, this plugin ends up in the `vendor` folder instead of `wp-content/mu-plugins`.

<details><summary>Repo info</summary>
Of course, to install this plugin, you need to add the repository to the composer.json file

```
...
"repositories": [
  {
    "type": "vcs",
    "url": "https://github.com/Automattic/Cron-Control"
   },
  ...
]
...
```
</details>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,7 @@
 {
   "name": "automattic/cron-control",
   "description": "Execute WordPress cron events in parallel, with custom event storage for high-volume cron.",
+  "type": "wordpress-muplugin",
   "license": "GPL-2.0",
   "require": {
     "php": ">=7.4.0"


rafaucau:
Another option is to release an official update on https://wordpress.org/plugins/wp-cron-control/ so that the latest version appears on WP Packagist: https://wpackagist.org/search?q=wp-cron-control&type=any&search=. Currently, the plugin has not been updated for 6 years.

ethitter:
The "WP-Cron Control" plugin on WordPress.org is not related to this plugin. They're similarly named because I'm not great at naming things, and although both deal with cron and I contributed to both, they're otherwise unrelated. The repo for the other one is archived here: https://github.com/Automattic/WP-Cron-Control; it's probably best that I close the plugin on WP.org as it's no longer maintained.

rafaucau:
Oh, this repo links to that plugin, so I thought it was just that new versions don't go to wp.org
image

rafaucau:
In any case, could you merge this? It's a change that will be very helpful, especially since this plugin is not available outside of GitHub.

rafaucau:
@WPprodigy

PR #375, opened 2024-02-07
Fix PHPUnit tests by updating the database health check

noahtallen:
Looking into the logs for the failing PHPUnit tests, I came across this GitHub issue, which describes an identical problem: https://github.com/MariaDB/mariadb-docker/issues/523. (Which is that the mariadb service fails to start, but doesn't have an error message.) The suggested fix was to update the health check command. I assume GitHub Actions is using the health command to detect if the service is running, and something about the old command must not work anymore.

That worked, but then I got a different error that PHPUnit Polyfills was out of date, so I also updated that to v1.1.

For testing instructions, just double check that the PHPUnit checks on this PR ran the test suite :)

### DIFF
--- a/.github/workflows/ci-php.yaml
+++ b/.github/workflows/ci-php.yaml
@@ -32,7 +32,7 @@ jobs:
           MYSQL_ROOT_PASSWORD: root
         ports:
           - 3306
-        options: --health-cmd="mysqladmin ping" --health-interval=5s --health-timeout=5s --health-retries=5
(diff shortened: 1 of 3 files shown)


WPprodigy (approved):

rinatkhaziev:
Where does healthcheck.sh live @noahtallen @WPprodigy ?

noahtallen:
It's part of the mariadb image, I believe: https://mariadb.com/kb/en/using-healthcheck-sh/

